PR #1, opened 2018-05-11
JSON RPC.

cgcardona:
Documenting the JSON RPC.


schancel:
Thanks for putting this up.  I'll give this a few days for review, and if nobody objects we can merge it.

PR #6, opened 2018-05-13
Developer portal

cgcardona:


schancel:
This is looking good.  Let's proceed in this direction, but the spec and workgroups should be split up so that appropriate content can be given to the BCTF or C2 or whatever we want to call it.  There's the code-of-conduct, the ettiquette, and some other information about the workgroups that need to be visible.

cgcardona:
Sounds good. I'll look in to that.

schancel:
Not sure where the merge conflict came from?

cgcardona:
I'm not seeing a merge conflict.

schancel:
@brad1121 You want to merge this before or after your header changes?

schancel:
What command did you run?

image

brad1121:
@schancel Do this before, I can see there are going to be some conflicts which I think will be easier to sort out from my side than this one.

brad1121:
The new pages will also need to be changed to fit the new header format/layout.

PR #10, opened 2018-05-14
Add `opdatasigverify.md` specification

schancel:
Add opdatasigverify spec for November.

@fyookball @Mengerian @gandrewstone

### DIFF
--- a/spec/opdatasigverify.md
+++ b/spec/opdatasigverify.md
@@ -1,0 +1,126 @@
+---
+layout: specification
+title: Enable oracle-based data import via OP_DATASIGVERIFY
+date: 2018-02-22
+activation: 1542300000
+version: 0.2
+---
+
+## Introduction
+
+OP_DATASIGVERIFY allows signed data to be imported into a script.  This data can then have many uses, depending on the rest of the script, such as deciding spendability of several possible addresses.  This opcode therefore enables the powerful blockchain concept of an "oracle" -- an entity that publishes authoritative statements about extra-blockchain events -- to be used in the Bitcoin Cash blockchain.  For an example use of how this opcode can be used to enable binary contracts on any security or betting on any quantitatively decidable event (such as a sports match) please [click here](https://medium.com/@g.andrew.stone/bitcoin-scripting-applications-decision-based-spending-8e7b93d7bdb9).  But this is just one example; as the Bitcoin Cash Script language grows in expressiveness, it is anticipated that this opcode will be used in many other applications.
+
+## Requirements on miners, full nodes, and clients
+
+Miners, full nodes, and clients will implement the OP_DATASIGVERIFY opcode and activate it during the May 2018 hard fork.
+
+## OP_DATASIGVERIFY Specification
+
+*OP_DATASIGVERIFY uses a new opcode number*
+
+    Opcode (decimal): 187
+    Opcode (hex): 0xbb
+
+### Implementation details:
+When OP_DATASIGVERIFY is executed, the stack should look like:
+
+*top of stack*
+* *pubkeyhash*
+* *type and signature*
+* *data*
+
+If there are less then 3 items on the stack, the script fails.  If the pubkeyhash field is not 20 bytes, the script fails.  If the *'type and signature'* field is not 66 bytes, the script fails.  If the last byte of the *'type and signature'* field is not DATASIG_COMPACT_ECDSA (1), the script fails.
+
+The last byte of the *'type and signature'* (stack top - 1) field defines the signature type, and previous bytes are the actual signature.  Note that this format is different from the OP_CHECKSIG format -- its SIGHASH flag byte has no relationship to this signature type byte.
+
+OP_DATASIGVERIFY looks at this signature type byte to determine the signature validation algorithm and then executes that algorithm.  There is currently one defined algorithm:
+
+#### value 1: DATASIG_COMPACT_ECDSA
+
+This algorithm is the same operation as the Bitcoin Cash message verification RPC call (e.g signmessage/verifymessage).  The signature must therefore be a pubkey-recoverable signature encoded in bitcoin's "compact" format.  The authoritative specification of this signature is beyond the scope of this document, but in essence it is comprised of a 1 byte recovery ID, 32 bytes "r", 32 bytes "s".
+
+The algorithm first computes the double-sha256 hash of the byte string "Bitcoin Signed Message:\n" prepended to the supplied *data* (stack top - 2).  It then computes a pubkey from this hash and the provided signature (stack top - 1, without the first byte).  If the pubkey cannot be recovered, the script fails.  It then compares the hash160 (same as OP_HASH160) of the recovered pubkey to the provided *pubkeyhash* (stack top).  If the comparison fails, the script fails.
+
+Otherwise, the top 2 items are popped off the stack (leaving *data* on the top of the stack), and the opcode succeeds.
+
+## Examples
+
+* Simplest DATASIGVERIFY:
+  * output script:  `pubkeyhash OP_DATASIGVERIFY`
+  * spend script: `data sig`
+
+* Realistic example: Verify that `data` is signed by `dpubkeyhash` (signature `daddrsig`) and equals `x`.  Also do normal p2pkh transaction verification on `pubkeyhash` (with signature `txSig` and public key `pubkey`):
+  * output script: `dpubkeyhash OP_DATASIGVERIFY x OP_EQUALVERIFY OP_DUP OP_HASH160 pubkeyhash OP_EQUALVERIFY OP_CHECKSIGVERIFY
+  * spend script: `txSig pubkey data daddrsig`
+
+## Discussion
+
+### Why is the non-verify version unnecessary?
+
+OP_DATASIGVERIFY fails the script if signature validation fails.  But one could imagine another opcode, let's call it OP_CHECKDATASIG, that would instead push true or false onto the stack depending on whether signature validation suceeded or failed.  Therefore additional opcode would relate to OP_DATASIGVERIFY in the same way OP_CHECKSIG is to OP_CHECKSIGVERIFY.
+
+But, this instruction is not necessary because a Bitcoin Cash script is not exactly like a normal program.  Its fundamental and only purpose is to encumber spendability -- your input script must meet the output script's requirements in order to spend the coins.  Therefore, any program sequence that does not narrow the group of possible spenders is equivalent to a simple choice.  In the OP_CHECKDATASIG case, any potential spender can create a spend script that fails OP_CHECKDATASIG by pushing random bytes rather than a valid signature, so the "false" result is equivalent to (but less efficient than) an if statement on a pushed constant.
+
+Formally, any script with encumberances x, y, z and solutions x', y', and z', of the form:
+
+`pubkeyhash OP_CHECKDATASIG OP_IF x OP_ELSE y OP_ENDIF z`
+
+if-branch spend: `z' x' data sig`
+
+else-branch spend: `z' y' data badsig`
+
+Can be replaced by the script:
+
+`OP_IF pubkeyhash OP_DATASIGVERIFY x OP_ELSE y OP_ENDIF z`
+
+if-branch spend:  `z' x' data sig 1`
+
+else-branch spend: `z' y' 0`
+
+
+## Reference Implementation
+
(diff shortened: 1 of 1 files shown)


zquestz (approved):

spec/opdatasigverify.md
```diff
@@ -0,0 +1,126 @@
...
+version: 0.2
+---
+
+## Introduction
+
+OP_DATASIGVERIFY allows signed data to be imported into a script.  This data can then have many uses, depending on the rest of the script, such as deciding spendability of several possible addresses.  This opcode therefore enables the powerful blockchain concept of an "oracle" -- an entity that publishes authoritative statements about extra-blockchain events -- to be used in the Bitcoin Cash blockchain.  For an example use of how this opcode can be used to enable binary contracts on any security or betting on any quantitatively decidable event (such as a sports match) please [click here](https://medium.com/@g.andrew.stone/bitcoin-scripting-applications-decision-based-spending-8e7b93d7bdb9).  But this is just one example; as the Bitcoin Cash Script language grows in expressiveness, it is anticipated that this opcode will be used in many other applications.
+
+## Requirements on miners, full nodes, and clients
+
+Miners, full nodes, and clients will implement the OP_DATASIGVERIFY opcode and activate it during the May 2018 hard fork.
```
sickpig: what about:

spec/opdatasigverify.md
```diff
@@ -0,0 +1,126 @@
...
+`OP_IF pubkeyhash OP_DATASIGVERIFY x OP_ELSE y OP_ENDIF z`
+
+if-branch spend:  `z' x' data sig 1`
+
+else-branch spend: `z' y' 0`
+
+
+## Reference Implementation
+
+Please refer to [this github branch](https://github.com/gandrewstone/BitcoinUnlimited/tree/op_datasigverify) for a complete implementation.  Implementation is in src/test/interpreter.cpp/h and unit tests are located at src/test/script_tests.cpp.
```
sickpig: what about:

fyookball:
The reasoning not to have a separate OP_CHECKDATASIG seems to make sense because there's no practical use case to check the data and then not do anything with the evaluation.  In practice, OP_CHECKSIG is always combined with a verification op_code afaik.

spec/opdatasigverify.md
```diff
@@ -0,0 +1,126 @@
...
+If there are less then 3 items on the stack, the script fails.  If the pubkeyhash field is not 20 bytes, the script fails.  If the *'type and signature'* field is not 66 bytes, the script fails.  If the last byte of the *'type and signature'* field is not DATASIG_COMPACT_ECDSA (1), the script fails.
+
+The last byte of the *'type and signature'* (stack top - 1) field defines the signature type, and previous bytes are the actual signature.  Note that this format is different from the OP_CHECKSIG format -- its SIGHASH flag byte has no relationship to this signature type byte.
+
+OP_DATASIGVERIFY looks at this signature type byte to determine the signature validation algorithm and then executes that algorithm.  There is currently one defined algorithm:
+
+#### value 1: DATASIG_COMPACT_ECDSA
+
+This algorithm is the same operation as the Bitcoin Cash message verification RPC call (e.g signmessage/verifymessage).  The signature must therefore be a pubkey-recoverable signature encoded in bitcoin's "compact" format.  The authoritative specification of this signature is beyond the scope of this document, but in essence it is comprised of a 1 byte recovery ID, 32 bytes "r", 32 bytes "s".
```
deadalnix: Why does the signature must use a pubkey recoverable algorithm ? It doesn't seems like it's bringing anything to the table to bake in that constraint.

spec/opdatasigverify.md
```diff
@@ -0,0 +1,126 @@
...
+
+The algorithm first computes the double-sha256 hash of the byte string "Bitcoin Signed Message:\n" prepended to the supplied *data* (stack top - 2).  It then computes a pubkey from this hash and the provided signature (stack top - 1, without the first byte).  If the pubkey cannot be recovered, the script fails.  It then compares the hash160 (same as OP_HASH160) of the recovered pubkey to the provided *pubkeyhash* (stack top).  If the comparison fails, the script fails.
```
deadalnix: This part is inconsistent with other parts of the spec.

spec/opdatasigverify.md
```diff
@@ -0,0 +1,126 @@
...
+
+Otherwise, the top 2 items are popped off the stack (leaving *data* on the top of the stack), and the opcode succeeds.
+
+## Examples
+
+* Simplest DATASIGVERIFY:
+  * output script:  `pubkeyhash OP_DATASIGVERIFY`
```
deadalnix: It does't make sense to bake the pubkey hash algorithm into that opcode.

spec/opdatasigverify.md
```diff
@@ -0,0 +1,126 @@
...
+## Requirements on miners, full nodes, and clients
+
+Miners, full nodes, and clients will implement the OP_DATASIGVERIFY opcode and activate it during the May 2018 hard fork.
+
+## OP_DATASIGVERIFY Specification
+
+*OP_DATASIGVERIFY uses a new opcode number*
+
+    Opcode (decimal): 187
+    Opcode (hex): 0xbb
```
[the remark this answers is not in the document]
deadalnix: Namely, strict DER encoding.

deadalnix (changes requested):
This spec is quite not in shape it needs to. In addition to the inline comments, The spec is lacking impact of this on sigops count - clearly these ops should count as sigops.

spec/opdatasigverify.md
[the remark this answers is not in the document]
fyookball: is there at least one use-case for OP_DATACHECKSIG standalone?

spec/opdatasigverify.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Implementation details:
+When OP_DATASIGVERIFY is executed, the stack should look like:
+
+*top of stack*
+* *pubkeyhash*
+* *type and signature*
+* *data*
+
+If there are less then 3 items on the stack, the script fails.  If the pubkeyhash field is not 20 bytes, the script fails.  If the *'type and signature'* field is not 66 bytes, the script fails.  If the last byte of the *'type and signature'* field is not DATASIG_COMPACT_ECDSA (1), the script fails.
```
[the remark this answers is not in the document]
deadalnix: Namely, strict DER encoding.

spec/opdatasigverify.md
gandrewstone: So that the input script does not have to include the pubkey...

spec/opdatasigverify.md
gandrewstone: verses using the unhashed pubkey?  The reason I did it this way is that pubkey hash is shorter, it hides the pubkey until use, and it is aligned with how signmessage/verifymessage works meaning that these RPCs can be used to create the spend and redeem scripts.

spec/opdatasigverify.md
sickpig: I would also change `hard fork` with `protocol update`

spec/opdatasigverify.md
Mengerian: Has there been any discussion on the pros and cons of following the signmessage/verifymessage format?

In many ways, it feels like it would be more elegant to keep the implementation closer to OP_CHECKSIG. Just have the inputs be a Public key, data, and bare DER signature (maybe it would also make sense for the signature to include a "sighash type" similar to OP_CHECKSIG). If scripts don't want to reveal the Public Key in the output script, it can supply a hash in a similar manner as P2PKH.

Since it's consensus critical, it's unclear to me whether the signmessage/verifymessage format has completely well-defined behavior, whereas something closer to unmodified OP_CHECKSIG is battle tested.

spec/opdatasigverify.md
```diff
@@ -0,0 +1,126 @@
...
+
+The algorithm first computes the double-sha256 hash of the byte string "Bitcoin Signed Message:\n" prepended to the supplied *data* (stack top - 2).  It then computes a pubkey from this hash and the provided signature (stack top - 1, without the first byte).  If the pubkey cannot be recovered, the script fails.  It then compares the hash160 (same as OP_HASH160) of the recovered pubkey to the provided *pubkeyhash* (stack top).  If the comparison fails, the script fails.
+
+Otherwise, the top 2 items are popped off the stack (leaving *data* on the top of the stack), and the opcode succeeds.
+
+## Examples
+
+* Simplest DATASIGVERIFY:
+  * output script:  `pubkeyhash OP_DATASIGVERIFY`
+  * spend script: `data sig`
```
Mengerian: This example seems strange. It doesn't seem to make sense for the ScriptSig to supply both the data and the signature.

spec/opdatasigverify.md
```diff
@@ -0,0 +1,126 @@
...
+    Opcode (decimal): 187
+    Opcode (hex): 0xbb
+
+### Implementation details:
+When OP_DATASIGVERIFY is executed, the stack should look like:
+
+*top of stack*
+* *pubkeyhash*
+* *type and signature*
+* *data*
```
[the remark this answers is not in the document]
gandrewstone: The data is unknown when the output script is created.  It is provided by the input script and the signature verifies that the oracle produced this data rather than it being invented by the input script author.  The hash_of_data proposal is not applicable here -- the purpose of this opcode is not to make a commitment to known but unrevealed data.  Think stock price or sporting match scores...

gandrewstone:
@schancel I am happy to make some of the requested changes based on the above discussion, but since you are the PR author I'm not sure whether I should or you want to.  Let me know and if so, how do you want me to make them?

sickpig:
paging @schancel 

what do you think about @gandrewstone's proposal made here

https://github.com/bitcoincashorg/bitcoincash.org/pull/10#issuecomment-393272723

spec/opdatasigverify.md
Mengerian: Since OP_DATASIGVERIFY leaves the data on the Stack on completion, if the data is a value of zero, and is signed properly, then OP_DATASIGVERIFY would complete successfully, but the script in this example would fail since it ends with a zero value left on the stack.

Not sure if this is a practical problem, but it seems like a strange edge case.

ClemensLey:
My two cents: Like @Mengerian I think that the inputs to OP_CHECKDATASIG should work as similar as possible to OP_CHECKSIG. This would mean that the parameters should be a public key, the message to be signed, and a signature. Like other op codes I think that it should pop the parameters from the stack and push the result (a boolean) onto the stack.

I think the op code would be easier to use if there is only a signature on the stack, not signature + sig_type. If we find strong use cases for other signature checking algorithm we could consider adding a signature type as a 4th parameter, but my feeling is that we should be fine with just a standard signature checking procedure.

Also I think it would be better if we could drop the requirement that the signature must be pubkey-recoverable.

All in all, I think that OP_CHECKDATASIG could be a valuable addition to the Bitcoin scripting language. We just need to do a little work to ensure that it is as easy to use as possible.

Mengerian:
This proposal has been superseded by https://github.com/bitcoincashorg/bitcoincash.org/pull/93

spec/opdatasigverify.md
```diff
@@ -0,0 +1,126 @@
...
+layout: specification
+title: Enable oracle-based data import via OP_DATASIGVERIFY
+date: 2018-02-22
+activation: 1542300000
+version: 0.2
+---
+
+## Introduction
+
+OP_DATASIGVERIFY allows signed data to be imported into a script.  This data can then have many uses, depending on the rest of the script, such as deciding spendability of several possible addresses.  This opcode therefore enables the powerful blockchain concept of an "oracle" -- an entity that publishes authoritative statements about extra-blockchain events -- to be used in the Bitcoin Cash blockchain.  For an example use of how this opcode can be used to enable binary contracts on any security or betting on any quantitatively decidable event (such as a sports match) please [click here](https://medium.com/@g.andrew.stone/bitcoin-scripting-applications-decision-based-spending-8e7b93d7bdb9).  But this is just one example; as the Bitcoin Cash Script language grows in expressiveness, it is anticipated that this opcode will be used in many other applications.
```
baby636, line 11: https://medium.com/@g.andrew.stone/bitcoin-scripting-applications-decision-based-spending-8e7b93d7bdb9

baby636 (commented):
https://medium.com/@g.andrew.stone/bitcoin-scripting-applications-decision-based-spending-8e7b93d7bdb9

PR #23, opened 2018-05-22
Split menus into primary & sub parts.  all pages use the same template.

brad1121:
Modified the Rakefile which got it working for me, including translations.
Works with Make also.


Rakefile
```diff
@@ -1,6 +1,14 @@
 require 'bundler'
 Bundler.setup
 
+namespace :docker do
+  desc "build docker images"
+  task :build => ['translations:build'] do
```
schancel: This stuff is entirely redundant now i believe.  No need for translations or anything else.

_config.yml
```diff
@@ -47,28 +47,28 @@ exclude:
...
-#  - pl
-#  - pt-BR
-#  - pt-PT
-#  - ru
-#  - sk
-#  - tr
-#  - vi
-#  - zh-CN
-#  - zh-TW
+# - cs
```
schancel: We should figure out if the jekyll plugin has a way to only render some languages in a dev mode.

index.html
```diff
@@ -1,8 +1,8 @@
 ---
 layout: page
 ---
-
-{% include home/menu.html %}
+{% assign isHome = true %}
```
schancel: This can go up above.

```
---
layout: page
isHome: true
---
```

schancel: You'll need to use `page.isHome` elsewhere rather than just `isHome`

Rakefile
brad1121: Whilst it is redundant when using Make, if you use Rake to build, it fails. At least for me this is the case.
I can revert back, and update the readme to build with Make not Rake.

schancel: Yes, we should just remove Rake entirely, and the rakefile.

schancel:
Let's do some of this here: https://github.com/bitcoincashorg/bitcoincash.org/pull/26

brad1121:
Next round of updates. This should be right to go.

Added the new pages into the menu system.
Set the new pages up to follow the new header + submenu layout.

_config-dev.yml
```diff
@@ -47,28 +47,28 @@ exclude:
 
 languages:
  - en
-#  - cs
```
schancel: Why did this get changed?

_i18n/en.yml
```diff
@@ -107,6 +107,7 @@ workgroups:
   workgroups:
     description: "The following is a topical list of Bitcoin Cash Development Workgroups.  Each group should be focused around a specific subject.  They may exist for a limited time, or be of ongoing interest."
     sections:
+
```
schancel: Wild newline

index.html
```diff
@@ -1,9 +1,9 @@
 ---
 layout: page
+isHome: true
```
schancel: It's too bad there isn't a better way to do this.  @PorlyBe Do you have any idea how to detect the current path of the page in jekyll?

learn.html
```diff
@@ -1,8 +1,8 @@
 ---
-layout: base
+layout: page
 use-site-title: true
```
schancel: This is specific to a theme we're not using.  We should remove it.

schancel:
* Some of the menu links don't work
* The workgroups being at the same level as the dev portal is odd
* If you goto workgroups you get "developers" : http://127.0.0.1:4000/workgroups.html
* If you goto spec, you get "workgroups" and specifications: http://127.0.0.1:4000/specs.html

These should be fixed before merging.

PR #25, opened 2018-05-22
update documentation of NUM2BIN to match implementation

Danconnolly:
A discrepancy has been found between the documentation and the implementation of the NUM2BIN opcode. This updates the documentation to make the implementation.

### DIFF
--- a/spec/may-2018-reenabled-opcodes.md
+++ b/spec/may-2018-reenabled-opcodes.md
@@ -3,7 +3,7 @@ layout: specification
 title: Restore disabled script opcodes, May 2018
 date: 2018-04-05
 activation: 1526400000
-version: 0.3
+version: 0.4
 ---
 
 ## Introduction
@@ -333,9 +333,10 @@ Unit tests:
 Convert the numeric value into a byte sequence of a certain size, taking account of the sign bit.
 The byte sequence produced uses the little-endian encoding.
 
-    `n m OP_NUM2BIN -> x`
+    n m OP_NUM2BIN -> x
     
-    where m and n are interpreted as numeric values
+where `m` and `n` are interpreted as numeric values. `m` must be minimally encoded and <= 4 bytes long.
+`n` can be up to `MAX_SCRIPT_ELEMENT_SIZE` long and does not need to be minimally encoded.
 
 See also `OP_BIN2NUM`.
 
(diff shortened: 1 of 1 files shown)


spec/may-2018-reenabled-opcodes.md
```diff
@@ -333,9 +333,10 @@ Unit tests:
 Convert the numeric value into a byte sequence of a certain size, taking account of the sign bit.
 The byte sequence produced uses the little-endian encoding.
 
-    `n m OP_NUM2BIN -> x`
+    n m OP_NUM2BIN -> x
     
-    where m and n are interpreted as numeric values
+where `m` and `n` are interpreted as numeric values. `m` must be minimally encoded and <= 4 bytes long.
```
jasonbcox: This seems nit-picky, but can you swap `m` and `n`? I thought the minimal encoding was referring to the first operand due to the ordering in the first sentence, despite the ordering be n m in the definition above.

spec/may-2018-reenabled-opcodes.md
```diff
@@ -3,7 +3,7 @@ layout: specification
 title: Restore disabled script opcodes, May 2018
 date: 2018-04-05
```
schancel: Is this the original date of publication, or are we updating it with the versions?

spec/may-2018-reenabled-opcodes.md
schancel: We should define somewhere what it means to interpret something as a `numeric value`

spec/may-2018-reenabled-opcodes.md
```diff
@@ -369,7 +369,7 @@ Unit tests:
 
 Convert the byte sequence into a numeric value, including minimal encoding. The byte sequence must encode the value in little-endian encoding.
 
-    `x1 OP_BIN2NUM -> n`
+    x1 OP_BIN2NUM -> n
```
schancel: Why remove the literal styling?

spec/may-2018-reenabled-opcodes.md
Danconnolly: I left it as the original date of publication, added an "updated" field.

spec/may-2018-reenabled-opcodes.md
Danconnolly: "code" style is already applied by the indent. the back-quotes appear in the final text.

spec/may-2018-reenabled-opcodes.md
Danconnolly: yes. originally, that was supposed to be covered by the "Script data types" section above.

Danconnolly: I went a step further and renamed them to "a" and "b", to be consistent with other parts of the document, although other parts could do with a review too. I also find "a" and "b" easier to differentiate than "m" and "n", but maybe thats just me.

jasonbcox:
LGTM

PR #27, opened 2018-05-28
Fix corrupt link for May-2018 hardfork

muojp:
All links for `https://github.com/bitcoincashorg/spec/blob/master/may-2018-hardfork.md` should be replaced by `https://github.com/bitcoincashorg/bitcoincash.org/blob/master/spec/may-2018-hardfork.md` due to recent repo reorganization.

Of course, getting rid of the entire hardfork announcement would be one way. But as a quick fix, I propose pointing at correct links because I'm sure many people are getting 404 error when Googling official info (incl. me). 🚶


schancel:
Thank you for putting this up.  However, this has been superseded by removing the announcement altogether.

PR #33, opened 2018-05-30
Add new service

sviridenkovalera:
Add new service


sviridenkovalera:
how long to wait until accept PR?

BUYMERCIER:
Hi @sviridenkovalera. Please fill-in the [get listed form](https://www.bitcoincash.org/get-listed.html). Also, your service seems to promote an ICO which does not qualify for being listed here.

PR #38, opened 2018-06-05
Error 404 page added

BUYMERCIER:


nginx.conf
```diff
@@ -43,7 +44,7 @@ server {
     }
 
     rewrite ^/letter-from-the-ceo.pdf$ https://falkvinge.net/wp-content/uploads/2017/11/letter-from-the-ceo.pdf permanent;
-    
+
```
schancel: Try to check your patches for extra stuff.  You can stage just portions with `git add -p`

Issue #39, opened 2018-06-05
Fix gradient aliasing in grey Bitcoin Cash Accepted PNGs


maplesyrupsucker:
Revise pngs and zip for 7.

maplesyrupsucker:
![7-bitcoin-cash-accepted-here-full](https://user-images.githubusercontent.com/3764321/40957122-9e4ddef0-68ce-11e8-8191-6fba4eb89069.png)
![7-bitcoin-cash-accepted-here-large](https://user-images.githubusercontent.com/3764321/40957123-9e88f29c-68ce-11e8-8fb7-b12503979cd2.png)
![7-bitcoin-cash-accepted-here-medium](https://user-images.githubusercontent.com/3764321/40957124-9ec8ecee-68ce-11e8-8053-c1fc47c0a16f.png)
![7-bitcoin-cash-accepted-here-small](https://user-images.githubusercontent.com/3764321/40957126-9f086ab8-68ce-11e8-86d3-2439b91861ae.png)

PR #41, opened 2018-06-05
Revised raster png's of Grey BCH accepted here. 

maplesyrupsucker:
Revised SVG to include raster BG. Vector Aliasing was causing strange lines in PNG.

Re: #39 issue


maplesyrupsucker:
Good question - I'll be sure to remove that. Looks like maybe something from jekyll serve perhaps. Will remove that before merging

PR #43, opened 2018-06-05
Nginx languages

brad1121:
Added auto language support to nginx
Added language menu item
Cleaned up some other menu items


schancel:
This needs to be rebased.  Can you please split this into two PRs?  Lang detection, and then the menu cleanup?

nginx.conf
```diff
@@ -35,6 +35,8 @@ server {
     server_name     bitcoincash.org;
     error_page 404  /404.html;
 
+
```
schancel: There's some extra newlines in this file.



.

PR #44, opened 2018-06-07
Fixed tense as May 15th upgrade is past

BUYMERCIER:


schancel:
These blurbs should just be removed from the site altogether.

PR #46, opened 2018-06-07
Adding CoinSwitch in exchange list

vimalsagar:


vimalsagar:
By when will this change go live?

PR #49, opened 2018-06-12
Translations re-enabled

BUYMERCIER:
Translations now working with the Jekyll Website

### DIFF
--- a/_config-dev.yml
+++ b/_config-dev.yml
@@ -1,4 +1,4 @@
-baseurl: 
+baseurl:
 
 baseurl: ""
 name: BitoinCash.org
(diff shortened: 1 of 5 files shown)


_config-dev.yml
```diff
@@ -1,4 +1,4 @@
-baseurl:
```
BUYMERCIER: Removed trailing space

_config-dev.yml
```diff
@@ -47,30 +47,30 @@ exclude:
...
-# - pt-BR
-# - pt-PT
-# - ru
-# - sk
-# - tr
-# - vi
-# - zh-CN
-# - zh-TW
+ - cs
+ - de-ch
```
schancel: These should be re-enabled in prod, not in dev.  You can run the prod environment locally if you need to.

schancel:
This should be two pull requests, one to fix the menu, and one to fix the language redirects.

PR #54, opened 2018-06-14
Language redirection fixed

BUYMERCIER:
Translation redirected to bitcoincash.org/fr instead of bitcoincash.org/<uri>/fr

### DIFF
--- a/_includes/main-menu.html
+++ b/_includes/main-menu.html
@@ -73,7 +73,7 @@
                                     <div class="dropdown__content lang-menu">
                                         <ul class="menu-vertical">
                                             {% for item in site.data.languages[] %}
-                                                <li><a class="external" href="{% unless item[0] == 'en' %}{{ item[0] | prepend: site.baseurl_root }}{% endunless %}">{{ item[1] }}</a></li>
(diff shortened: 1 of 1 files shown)


_includes/main-menu.html
```diff
@@ -73,7 +73,7 @@
                                     <div class="dropdown__content lang-menu">
                                         <ul class="menu-vertical">
                                             {% for item in site.data.languages[] %}
-                                                <li><a class="external" href="{% unless item[0] == 'en' %}{{ item[0] | prepend: site.baseurl_root }}{% endunless %}">{{ item[1] }}</a></li>
+                                                <li><a class="external" href="/{% unless item[0] == 'en' %}{{ item[0] | prepend: site.baseurl_root }}{% endunless %}">{{ item[1] }}</a></li>
```
schancel: This is what `prepend: site.base_url` does.  The real problem is here:

https://github.com/bitcoincashorg/bitcoincash.org/blob/master/_config-prod.yml#L1

it should likely be "/"

schancel: This will work for now.  The unless block means english ends up broken.

PR #58, opened 2018-06-20
Added get listed / submit a listing button, added form page

maplesyrupsucker:


_includes/home/nodes.html
```diff
@@ -23,7 +23,7 @@ <h2>{% t 'headings.nodes' %}</h2>
                 <a href="https://www.bitprim.org/" target="_blank"><img alt="Bitprim" src="/img/wallets/bitprim.png"></a>
             </div>
             <div class="col-sm-12 text-center">
-                <a href="/get-listed.html" class="btn btn-default">Submit a listing</a>
+                <a href="/get-listed.html" class="btn btn-default">Submit a listing request</a>
```
schancel: This need's to be i18n-ified.  Also, where's the actual get-listed page?

maplesyrupsucker: Fixed - get-listed.html is in dir root with index.html

get-listed.html
```diff
@@ -0,0 +1,8 @@
+<html> 
+<head> 
+  <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0"> 
+  <title>{% t 'headings.listingrequest' %}</title>
```
BUYMERCIER: Title doesn't display correctly

maplesyrupsucker:
This pull includes a fix for newly found bug preventing title tags i18n from working properly. 

Get listed form setup in branded page - to setup multi-lang we'll need conditions setup via form editor for users to select lang. That or we limit submissions to english only, which would be ideal since reviewers of content for the most part speak/read english.

image

merc1er (approved):

PR #60, opened 2018-06-22
Moved secondary nav out of primary header position below header. 

maplesyrupsucker:
Adds visual hierarchy to nav items

image


maplesyrupsucker:
Revised secondary nav to keep it in container on big screens

Issue #62, opened 2018-06-26
Redirect problem possibly related with languages with variants


m4ktub:
When I request http://www.bitcoincash.org/ with a browser that sends `ACCEPT_LANGUAGE: pt-PT,pt;q=0.9,en-US;q=0.8,en;q=0.7` I get constant redirects.

- http://www.bitcoincash.org/
- https://www.bitcoincash.org/
- http://www.bitcoincash.org/pt-PT/
- https://www.bitcoincash.org/pt-PT/
- http://www.bitcoincash.org/pt-PT/pt-PT/
- https://www.bitcoincash.org/pt-PT/pt-PT/
- and so and so forth.

I could reproduce that behavior with `zh-CN` but not `ko`. It seems to manifest only when the language includes a variant. Nevertheless, when using `ko`, the unexpected and unecessary redirect from `https` back to `http` still occurs.

- http://www.bitcoincash.org/
- https://www.bitcoincash.org/
- http://www.bitcoincash.org/ko/ :thinking: 
- https://www.bitcoincash.org/ko/

PR #63, opened 2018-06-28
Language auto redirection disabled

BUYMERCIER:
Issue #62 

Language auto redirections temporarily disabled to avoid infinite looping on certain languages.


schancel:
Not the way to handle this.  Pushing an commit.

PR #64, opened 2018-06-29
height override on submission form - fixes ipad 1px height bug

maplesyrupsucker:
resolves 1px height bug on ipads viewing form

### DIFF
--- a/css/graphics.css
+++ b/css/graphics.css
@@ -1387,9 +1387,6 @@ flex-wrap: wrap;}
 
 
 }
-
-
-
 .type--fine-print {
     font-size: 0.85714286em;
 }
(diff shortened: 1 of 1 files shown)


css/graphics.css
```diff
@@ -1382,6 +1387,9 @@ flex-wrap: wrap;}
 
 
 }
+
+
```
BUYMERCIER: Why so many line breaks?

css/graphics.css
```diff
@@ -1382,6 +1387,9 @@ flex-wrap: wrap;}
...
+
```
deadalnix: ???

maplesyrupsucker: resolved in latest commit

PR #67, opened 2018-07-03
Merged jonalds english in, resolved conflicts

maplesyrupsucker:


maplesyrupsucker:
Included a revision that should allow us to go live with the new english copy without breaking any other langs. When the new translations are added the new templates will be included.

PR #92, opened 2018-08-04
BCH spec for Block and Transaction

lparis:
@schancel @corbinf Content is ready to be merged, headers need to be fixed for presentation, thanks

### DIFF
--- a/spec/block.md
+++ b/spec/block.md
@@ -1,15 +1,18 @@
-+++
-date = "2018-07-05"
-title = "Block"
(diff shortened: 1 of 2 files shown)


spec/block.md
```diff
@@ -0,0 +1,142 @@
++++
```
sickpig: correct syntax to define headers is triple dash. Probably you know it already since you mentions that header needs to be fixed in the PR description, just want to make sure to not miss it before publishing.

lparis:
@sickpig Great, thanks.

PR #93, opened 2018-08-06
Specification for OP_CHECKDATASIG and OP_CHECKDATASIGVERIFY

Mengerian:
This should supersede https://github.com/bitcoincashorg/bitcoincash.org/pull/10

It is a modified version of Andrew Stone's original proposal based on review feedback.

Thanks to Andrew Stone for the original proposal, and thanks to the reviewers:
 - Clemens Ley
 - Chris Pacia
 - Amaury Sechet
 - Andrea Suisani
 - Tom Harding

### DIFF
--- a/spec/op_checkdatasig.md
+++ b/spec/op_checkdatasig.md
@@ -25,7 +25,7 @@ OP_CHECKDATASIG fails immediately if the stack is not well formed. To be well fo
   * `<msg>` can be any string
   * `<sig>` must follow the strict DER encoding as described in [2] and the S-value of `<sig>` must be at most the curve order divided by 2 as described in [3]
 
-If the stack is well formed, then OP_CHECKDATASIG pops the top three elements [`<sig>`, `<msg>`, `<pubKey>`] from the stack and pushes true onto the stack if the signature is valid with respect to the single-SHA256 hash of the message and the public key using the secp256k1 elliptic curve. Otherwise, it pops three elements and pushes false onto the stack in the case that `<sig>` is the empty string and fails in all other cases.
+If the stack is well formed, then OP_CHECKDATASIG pops the top three elements [`<sig>`, `<msg>`, `<pubKey>`] from the stack and pushes true onto the stack if the signature is valid with respect to the raw single-SHA256 hash (no further padding, serialization, length etc. is concatenated to the message bytes from the stack) of the message and the public key using the secp256k1 elliptic curve. Otherwise, it pops three elements and pushes false onto the stack in the case that `<sig>` is the empty string and fails in all other cases.
 
 Nullfail is enforced the same as for OP_CHECKSIG [3]. If the signature does not match the supplied public key and message hash, and the signature is not an empty byte array, the entire script fails.
 
--- a/spec/op_checkdatasig.md
+++ b/spec/op_checkdatasig.md
@@ -9,7 +9,7 @@ version: 0.1
 OP_CHECKDATASIG
 ===============
 
-OP_CHECKDATASIG and OP_CHECKDATASIGVERIFY check whether a signature is valid with respect to a message hash and and a public key. The message hash is restricted to exactly 256 bits.
+OP_CHECKDATASIG and OP_CHECKDATASIGVERIFY check whether a signature is valid with respect to a message hash and a public key. The message hash is restricted to exactly 256 bits.
 
 OP_CHECKDATASIG permits data to be imported into a script, and have its validity checked against some signing authority such as an "Oracle".
 
@@ -162,6 +162,6 @@ References
 
 [5] [Andrew Stone’s OP_DATASIGVERIFY](https://github.com/BitcoinUnlimited/BitcoinUnlimited/blob/bucash1.3.0.0/doc/opdatasigverify.md)
 
-[6] [Andrew Stone article on Scripting](https://medium.com/@g.andrew.stone/bitcoin-scripting-applications-decision-based-spending-8e7b93d7bdb9)
+[6] [Andrew Stone's article on Scripting](https://medium.com/@g.andrew.stone/bitcoin-scripting-applications-decision-based-spending-8e7b93d7bdb9)
 
-[7] [Peer Review of Stone Proposal](https://github.com/bitcoincashorg/bitcoincash.org/pull/10)
+[7] [Peer Review of Andrew Stone's Proposal](https://github.com/bitcoincashorg/bitcoincash.org/pull/10)
--- a/spec/op_checkdatasig.md
+++ b/spec/op_checkdatasig.md
@@ -3,7 +3,7 @@ layout: specification
 title: OP_CHECKDATASIG and OP_CHECKDATASIGVERIFY Specification
 date: 2018-08-12
 activation: 1542300000
-version: 0.5
+version: 0.6
 ---
 
 OP_CHECKDATASIG
@@ -138,6 +138,39 @@ Sample Implementation [4, 5]
                             }
                         }
                     } break;
+```
+
+Sample Usage
+------------
+
+The following example shows a spend and redeem script for a basic use of CHECKDATASIG.  This example validates the signature of some data, provides a placeholder where you would then process that data, and finally allows one of 2 signatures to spend based on the outcome of the data processing.
+
+### spend script:
+```
+push txsignature
+push txpubkey
+push msg
+push sig
+```
+### redeem script:
+```
+                                (txsig, txpubkey msg, sig)
+OP_OVER                         (txsig, txpubkey, msg, sig, msg)
+push data pubkey                (txsig, txpubkey, msg, sig, msg, pubkey)
+OP_CHECKDATASIGVERIFY           (txsig, txpubkey, msg)
+```
+Now that msg is on the stack top, the script can write predicates on it,
+resulting in the message being consumed and a true/false condition left on the stack: (txpubkey, txsig, boolean)
+```
+OP_IF                           (txsig, txpubkey)
+  OP_DUP                        (txsig, txpubkey, txpubkey)
+  OP_HASH160                    (txsig, txpubkey, address)
+  push <p2pkh spend address>    (txsig, txpubkey, address, p2pkh spend address)
+  OP_EQUALVERIFY                (txsig, txpubkey)
+  OP_CHECKSIG
+OP_ELSE
+  (same as if clause but a different <p2pkh spend address>)
+OP_ENDIF
 ```
 
 History
--- a/spec/op_checkdatasig.md
+++ b/spec/op_checkdatasig.md
@@ -25,7 +25,7 @@ OP_CHECKDATASIG fails immediately if the stack is not well formed. To be well fo
   * `<msg>` can be any string
   * `<sig>` must follow the strict DER encoding as described in [2] and the S-value of `<sig>` must be at most the curve order divided by 2 as described in [3]
 
-If the stack is well formed, then OP_CHECKDATASIG pops the top three elements [`<sig>`, `<msg>`, `<pubKey>`] from the stack and pushes true onto the stack if the signature is valid with respect to the raw single-SHA256 hash (no further padding, serialization, length etc. is concatenated to the message bytes from the stack) of the message and the public key using the secp256k1 elliptic curve. Otherwise, it pops three elements and pushes false onto the stack in the case that `<sig>` is the empty string and fails in all other cases.
+If the stack is well formed, then OP_CHECKDATASIG pops the top three elements [`<sig>`, `<msg>`, `<pubKey>`] from the stack and pushes true onto the stack if `<sig>` is valid with respect to the raw single-SHA256 hash (no further padding, serialization, length etc. is concatenated to the message bytes from the stack) of `<msg>` and `<pubKey>` using the secp256k1 elliptic curve. Otherwise, it pops three elements and pushes false onto the stack in the case that `<sig>` is the empty string and fails in all other cases.
 
 Nullfail is enforced the same as for OP_CHECKSIG [3]. If the signature does not match the supplied public key and message hash, and the signature is not an empty byte array, the entire script fails.
 
--- a/spec/op_checkdatasig.md
+++ b/spec/op_checkdatasig.md
@@ -1,3 +1,11 @@
+---
+layout: specification
+title: OP_CHECKDATASIG and OP_CHECKDATASIGVERIFY Specification
+date: 2018-08-06
+activation: 1542300000
+version: 0.1
+---
+
 OP_CHECKDATASIG
 ===============
 
--- a/spec/op_checkdatasig.md
+++ b/spec/op_checkdatasig.md
@@ -1,7 +1,7 @@
 ---
 layout: specification
 title: OP_CHECKDATASIG and OP_CHECKDATASIGVERIFY Specification
-date: 2018-08-12
+date: 2018-08-20
 activation: 1542300000
 version: 0.6
 ---
--- a/spec/op_checkdatasig.md
+++ b/spec/op_checkdatasig.md
@@ -3,7 +3,7 @@ layout: specification
 title: OP_CHECKDATASIG and OP_CHECKDATASIGVERIFY Specification
 date: 2018-08-12
 activation: 1542300000
-version: 0.3
+version: 0.4
 ---
 
 OP_CHECKDATASIG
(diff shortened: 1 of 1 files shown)


spec/op_checkdatasig.md
```diff
@@ -0,0 +1,167 @@
...
+
+[3] [Low-S and Nullfail Specification](https://github.com/bitcoin/bips/blob/master/bip-0146.mediawiki)
+
+[4] [Bitcoin ABC implementation](https://reviews.bitcoinabc.org/D1621)
+
+[5] [Andrew Stone’s OP_DATASIGVERIFY](https://github.com/BitcoinUnlimited/BitcoinUnlimited/blob/bucash1.3.0.0/doc/opdatasigverify.md)
+
+[6] [Andrew Stone article on Scripting](https://medium.com/@g.andrew.stone/bitcoin-scripting-applications-decision-based-spending-8e7b93d7bdb9)
+
+[7] [Peer Review of Stone Proposal](https://github.com/bitcoincashorg/bitcoincash.org/pull/10)
```
sickpig: Stone's Proposal

jasonbcox: I think these references should be consistent. I recommend spelling out the name completely "Andrew Stone's". Same goes for reference [6], making sure it has the 's

sickpig: agreed

Mengerian:
OK, I made Andrew Stone's name consistent, and fixed the double "and"

markblundeberg:
Very nice, I like this approach. I am curious about something, what happens with the hashtype byte that normally gets appended onto the DER signature in a regular CHECKSIG? Am I obligated to include it, exclude it, or either is fine?

Mengerian:
Hi Mark, There is no "type byte", so you would not include it. The signature looks like an OP_CHECKSIG signature in terms of DER and all other rules, but without the type byte appended.

Mengerian:
Updated spec so that the opcode always hashes the input message.

awemany:
@Mengerian: Is this compatible with the use case that I foresee might be handy, a.k.a. "Zero Conf Insurance" to use what goes into SIGHASH_FORKID and then check for two different signatures of the same transaction output?

I suspect the double hashing might break this use case - or am I missing something?

markblundeberg:
@Mengerian  Regarding the recent change to having `message` rather than `message_hash`, I completely understand the desire to discourage the writing of easily abused smart contract setups with bad crypto.

That said, I'd like to point out that having **double**-SHA256 built-in seems unnecessarily limiting as opposed to using single-SHA256. Using a single SHA256 is sufficient for purposes and also allows some things [like an atomic trade of BCH for a GPG signature on a small message](https://www.reddit.com/r/btc/comments/96fxvy/op_checkdatasig_is_copying_blockstream_and_is/e4520xa/) (I can elaborate more on the precise protocol if the sketch I provided is not clear). That example may be a bit contrived, but it shows how in principle double-SHA256 could exclude some applications that single-SHA256 will not. Cheers!

awemany:
Preliminary code (needs review and updates to all the tests still!!) to change to just a single "round" of SHA256 hashing: https://github.com/awemany/BitcoinUnlimited/tree/checkdatasig-single-hash-infrastructure

@markblundeberg: As you are interested in this as well, could you have a look at this? Thank you :)

awemany:
Before we go with the above, let me remark here that the currently implemented approach in my branch needs further refinement. We use SER_HASH to serialize the data, which means it is prepended with a length byte. This does not allow to replace DATASIGVERIFY in the new variant with SWAP SHA256 SWAP DATASIGVERIFY which I think is desirable.

awemany:
Ok, I updated the above git branch to make it the **RAW** , **SINGLE** SHA256 of the message stack item, as this is the most flexible implementation that still does not allow exploits due to not-following  the EC signature standards. It also simplified the code a lot.

However, I wonder about other attacks along the lines of length extension or so. I am not a cryptographer and someone more knowledgeable then me here should review that aspect.

It is also much simplified and this also hopefully makes review easier. Now on to updating the test cases!

spec/op_checkdatasig.md
```diff
@@ -0,0 +1,163 @@
...
+
+OP_CHECKDATASIG Specification
+-----------------------------
+
+### Semantics
+
+OP_CHECKDATASIG fails immediately if the stack is not well formed. To be well formed, the stack must contain at least three elements [`<sig>`, `<msg>`, `<pubKey>`] in this order where `<pubKey>` is the top element and
+  * `<pubKey>` must be a validly encoded public key
+  * `<msg>` can be any string
+  * `<sig>` must follow the strict DER encoding as described in [2] and the S-value of `<sig>` must be at most the curve oder divided by 2 as described in [3]
```
schancel: `oder` -> `order` ?

awemany:
Further comment: The branch in https://github.com/awemany/BitcoinUnlimited/tree/checkdatasig-single-hash-infrastructure now contains all code and tests updated for CHECKDATASIG/-VERIFY with single hashing.

I will now work on a phabricator diff but encourage anyone interested to test and review this. @markblundeberg  ?

awemany:
Ok, as an update (I deleted my last message as I it was premature), the branch in 

https://github.com/awemany/BitcoinUnlimited/tree/checkdatasig-single-hash-infrastructure

on top of BitcoinABC should now pass and contains an update to all the tests of CHECKDATASIG.

This is the above-mentioned changeset to change from double-hashing a Bitcoin-serialized message to just single-hashing the stack value.

spec/op_checkdatasig.md
```diff
@@ -23,9 +23,9 @@ OP_CHECKDATASIG Specification
...
-  * `<sig>` must follow the strict DER encoding as described in [2] and the S-value of `<sig>` must be at most the curve oder divided by 2 as described in [3]
+  * `<sig>` must follow the strict DER encoding as described in [2] and the S-value of `<sig>` must be at most the curve order divided by 2 as described in [3]
 
-If the stack is well formed, then OP_CHECKDATASIG pops the top three elements [`<sig>`, `<msg>`, `<pubKey>`] from the stack and pushes true onto the stack if the signature is valid with respect to the double-SHA256 hash of the message and the public key using the secp256k1 elliptic curve. Otherwise, it pops three elements and pushes false onto the stack in the case that `<sig>` is the empty string and fails in all other cases.
+If the stack is well formed, then OP_CHECKDATASIG pops the top three elements [`<sig>`, `<msg>`, `<pubKey>`] from the stack and pushes true onto the stack if the signature is valid with respect to the single-SHA256 hash of the message and the public key using the secp256k1 elliptic curve. Otherwise, it pops three elements and pushes false onto the stack in the case that `<sig>` is the empty string and fails in all other cases.
```
awemany: My anger with deadalnix on the handling of this whole thing aside, I would make it explicit that it is the *raw* SHA256 hash of the message. As the spec and the code were out of sync before in this regard. This is why I think it is better to be explicit here. ("No further padding, serialization, length etc. is concatenated to the message bytes from the stack"). Something like that maybe?

Mengerian: OK, done

spec/op_checkdatasig.md
```diff
@@ -25,7 +25,7 @@ OP_CHECKDATASIG fails immediately if the stack is not well formed. To be well fo
   * `<msg>` can be any string
   * `<sig>` must follow the strict DER encoding as described in [2] and the S-value of `<sig>` must be at most the curve order divided by 2 as described in [3]
 
-If the stack is well formed, then OP_CHECKDATASIG pops the top three elements [`<sig>`, `<msg>`, `<pubKey>`] from the stack and pushes true onto the stack if the signature is valid with respect to the single-SHA256 hash of the message and the public key using the secp256k1 elliptic curve. Otherwise, it pops three elements and pushes false onto the stack in the case that `<sig>` is the empty string and fails in all other cases.
+If the stack is well formed, then OP_CHECKDATASIG pops the top three elements [`<sig>`, `<msg>`, `<pubKey>`] from the stack and pushes true onto the stack if the signature is valid with respect to the raw single-SHA256 hash (no further padding, serialization, length etc. is concatenated to the message bytes from the stack) of the message and the public key using the secp256k1 elliptic curve. Otherwise, it pops three elements and pushes false onto the stack in the case that `<sig>` is the empty string and fails in all other cases.
```
markblundeberg: I presume this means the `<msg>` byte array should be directly passed into a standard SHA256 function, however, it's my understanding SHA256 always involves length padding ([detail in the FIPS PDF](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf#page=18)). It seems like this thing about 'raw SHA256 without padding' could be misunderstood as asking to use a modified SHA256?

(also I would replace "the message" with "`<msg>`" for clarity)

Anyway this is getting to look like a pretty awesome protocol. If I have the time one day I want to write a gist and proof-of-concept on how you could hire someone to review and PGP-sign a short message, and finally perform an atomic swap of the BCH payment for their PGP signature. :-D

Mengerian: OK, I changed "the message" the <msg>

Happy to change wording around the SHA256 description also... Do you suggest I just remove everything in the brackets?

dgenr8:
As in the example, OP_OVER may find new popularity with this, as it will be required any time the message is needed by the spend script.  For that reason wouldn't it be better to leave the message on the stack?  It seems rare that you would only care that any random thing was signed.

spec/op_checkdatasig.md
markblundeberg: Oop, I missed this reply. One way to clarify it would be to specify the pseudocode `SHA256(msg)`. 

Regarding the parenthical, is there normally some extra padding/serialization/length that bitcoin adds before calling SHA256, in other situations?

spec/op_checkdatasig.md
```diff
@@ -0,0 +1,163 @@
...
+                        if (!CheckDataSignatureEncoding(vchSig, flags,
+                                                        serror) ||
+                            !CheckPubKeyEncoding(vchPubKey, flags, serror)) {
+                            // serror is set
+                            return false;
+                        }
+
+                        bool fSuccess = false;
+                        if (vchSig.size()) {
+                            CHashWriter ss(SER_GETHASH, 0);
```
awemany: The implementation section needs an upgrade as well for single hashing.

Mengerian: good catch, I will update the implementation section

Mengerian:
I made a couple changes:
 - updated implementation section to match change to single-SHA256 on the input
 - added the example use section suggested by Andrew Stone

Mengerian:
@dgenr8 Regarding leaving the message on the stack, this was something that was discussed quite a bit during the design/review. I think the reasoning was to make the opcode more generic for different use cases, not assume the oracle use case. You may be correct that we'll end up seeing a lot of OP_OVER.

If you're interested, I can dig for more details on the justification for that decision.

At this point, I'm inclined to leave it, since it was thought about by the previous reviewers, and even if they made the wrong decision it's not a fatal flaw.

sickpig:
@Mengerian 

thanks for the update to specification.

> If you're interested, I can dig for more details on the justification for that decision.

the problem here is that part of the design/review process hasn't been done in the most transparent way. You wouldn't have needed to dig for more details if the process had been different. 

Initial spec PR https://github.com/bitcoincashorg/bitcoincash.org/pull/10 was opened by @schancel on May 14th based on doc and code by @gandrewstone, the discussion there ended pretty soon after first round of review and with the [offer](https://github.com/bitcoincashorg/bitcoincash.org/pull/10#issuecomment-393272723) from @gandrewstone to fix the issues rised (May 30th).

Other than that nothing else happened on that PR thread a part from a [message](https://github.com/bitcoincashorg/bitcoincash.org/pull/10#issuecomment-395932684) from @ClemensLey.

The actual review process continued [somewhere else ](https://gist.github.com/Mengerian/d200f8b3a238a256642e498ca796e7c4), problem was that that location was unknown to the most till Aug 6th, which is that day this specification PR has been opened. 

Didn't mean to point finger to anyone just wanted to point out we could have done better by having the review/design done in the open since the beginning, so that we could have gathered more feedback and maybe avoided some last [minute](https://reviews.bitcoinabc.org/D1653) [changes](https://reviews.bitcoinabc.org/D1652)

dgenr8:
The spender also has to push msg and sig in the opposite order versus how they are consumed by CHECKDATASIG.

When both the spend and redeem script have to make tweaks, the opcode isn't quite right.

Mengerian:
> the problem here is that part of the design/review process hasn't been done in the most transparent way. You wouldn't have needed to dig for more details if the process had been different.

Yes, I'm sure the way I did it was not perfect. I just saw that the original proposal was not getting sufficient much review, so I reached out to several people directly, including Clemens Ley, Chris Pacia, and Tom Harding. I just reached out to people that I thought would have some specialized knowledge in Script and opcodes.

Tom made one suggestion at the time, which was that "it should follow existing CHECKSIG semantics where possible". This was in-line with what other reviewers were saying, and what ended up happening.

It wasn't my proposal, but I wanted to help get it enough review so that it could be merged into ABC. I reached out to people who I thought would be knowledgeable, and tried to involve people who seemed interested.

Yeah, I learned that in the future, if I ever do this again, it would be better to try to get the public spec out there earlier. But I also have to say, it seemed like no-one was that interested earlier, and I had to push to get reviews. Then when it's getting merged everyone gets interested. So it's hard to know the best way to resolve that.

sickpig:
@Mengerian 


> But I also have to say, it seemed like no-one was that interested earlier, and I had to push to get reviews. 

I noticed the lack of interest too and I really do thank you for what you did and what you are doing in trying to get the process moving forward. On the other hand I don't see how the lack of interest has something to do w/ moving the design/review on a different place. 

> Then when it's getting merged everyone gets interested. So it's hard to know the best way to resolve that.

Humans are strange creature indeed :smile: 

Having said that I think that people start to get interested not just because an implementation got merged in ABC (https://reviews.bitcoinabc.org/D1621) on the 6th of Aug, but also because the 7th of August is the day when this PR has been opened for public review (It wouldn't have been possible to it before that date). It could be also that people attention got stimulated by all the drama built up and accumulated since the end of activities on #10 (end of May) and the opening of #93.

I'm sure for example that @awemany's interest ((https://reviews.bitcoinabc.org/D1652)) in this new op code stemmed from an idea he had which involve the use of OP_CHECKDATASIG to increase 0-conf transaction reliability. He conceived his idea pretty recently and he realized that there were a problem (using of SER_HASH and double-sha256 rather than single raw sha256 of the msg) in the ABC implementation because of this and that's why he raised the [issue](https://github.com/bitcoincashorg/bitcoincash.org/pull/93#issuecomment-412637195) in this very thread.

Other people  could have started looking at it due to other kind of [initiative](https://www.reddit.com/r/btc/comments/96fxvy/op_checkdatasig_is_copying_blockstream_and_is/) but still even in this case the results wasn't that bad, in fact another suggested change was included (https://reviews.bitcoinabc.org/D1646), i.e. user is not expected to hash the `msg`  anymore, and as Amury said in the above diff it makes the implementation "safer".

dgenr8:
@Mengerian Open source is about making the most of the free contributions people choose to make. It doesn't do any good wish they did things differently.

Mengerian:
@dgenr8 You must be misinterpreting something I said.

I have no expectation or wish for people to do things differently.

Mengerian:
Hi All,

I'm thinking it may be time to merge this specification soon. Seems there are no major objections, the only remaining item is the order of inputs.. but I think that's not a show-stopper issue, especially since there are different opinions on the matter.

It also turns out that it's the same way it is implemented in the Elements Alpha "OP_CHECKSIGFROMSTACK":
https://github.com/ElementsProject/elements/blob/fcc652f45806f3ad18f6403052d3f0ec6e366cb7/src/script/interpreter.cpp#L1312

I'm embarrassed to say I wasn't even aware of this opcode until someone posted to Reddit a few weeks ago accusing us of copying it. But it seems interesting that the also came up with the same input order. This makes me think that it can't be totally wrong, and there may be some value in retaining compatibility with them.

Anyhow, if someone could do the review to approve it, then I can proceed to merge in the next few days.

markblundeberg:
Regarding leaving `msg` on the stack, so far with my ideas ([1](https://gist.github.com/markblundeberg/af59d7cd234cbdb14dcf9e00f0ea2c17), [2](https://gist.github.com/markblundeberg/bd28871548108fc66d958018b1bde085)) this would not have helped (if left on stack, I would have to use OP_DROP to get rid of it).

The reason is that I have a multi-component `msg = A|B|C` where each part A, B, C has different meaning and gets used differently. Rather than taking `msg` and trying to parse it into pieces, it is more convenient to have each of A,B,C given as separate inputs and then only OP_CAT them together when it comes time to check the signature.

dgenr8:
@markblundeberg The signed data should always be available and explicitly dropped if necessary. In two of your examples (which are very nice btw) CHECKDATASIG is the last instruction so DROP is not necessary.

PR #94, opened 2018-08-07
Add Specification for 2018-11-15 Upgrade

Mengerian:


spec/2018-nov-upgrade.md
```diff
@@ -0,0 +1,56 @@
...
+
+With the exception of the coinbase transaction, transactions within a block shall be sorted by Transaction ID in ascending order. The coinbase transaction shall be the first transaction in a block.
+
+## OpCodes
+
+New opcodes OP_CHECKDATASIG and OP_CHECKDATASIGVERIFY will be enabled as specified in [op_checkdatasig.md](op_checkdatasig.md) [2].
+
+## Minimum Transaction Size
+
+Transactions that are smaller than 100 bytes shall be considered invalid.
```
jasonbcox: Perhaps make a note as to why: "This protects against a Merkle tree vulnerability that allows attackers to spoof transactions against SPV wallets."  Something like that.

schancel:
Thank you so much for taking care of this @Mengerian

spec/2018-nov-upgrade.md
```diff
@@ -0,0 +1,56 @@
...
+
+Transactions shall be considered invalid if an opcode with number greater than 96 (hex encoding 0x60) appears in a scriptSig.
+
+## Automatic Replay Protection
+
+When the median time past [1] of the most recent 11 blocks (MTP-11) is less than UNIX timestamp 1557921600 (May 2019 upgrade) Bitcoin Cash full nodes MUST enforce the following rule:
+
+ * `forkid` [3] to be equal to 0.
+
+When the median time past [1] of the most recent 11 blocks (MTP-11) is greater than or equal to UNIX timestamp 1557921600 (May 2019 upgrade) Bitcoin Cash full nodes implementing the November 2018 consensus rules SHOULD enforce the following change:
```
zquestz: Does this mean that all wallets and exchanges must update their transaction signing code to use 0xFF0001 in May 2019?

spec/2018-nov-upgrade.md
```diff
@@ -0,0 +1,56 @@
...
+
+ * Update `forkid` [3] to be equal to 0xFF0001.  ForkIDs beginning with 0xFF will be reserved for future protocol upgrades.
+
+This particular consensus rule MUST NOT be implemented by Bitcoin Cash wallet software.
```
ryanxcharles: This section is missing a key sentence to explain the motivation. Your plan is presumably to "hard fork" back to forkId 0 in May 2019 so that wallets that follow the forks never have to change the forkId. Would be great to see that clarifying sentence placed in this section for all of us who develop wallets :)

spec/2018-nov-upgrade.md
```diff
@@ -0,0 +1,56 @@
...
+date: 2018-07-29
+activation: 1542300000
+version: 0.1
+---
+
+## Summary
+
+When the median time past [1] of the most recent 11 blocks (MTP-11) is greater than or equal to UNIX timestamp 1542300000, Bitcoin Cash will execute an upgrade of the network consensus rules according to this specification. Starting from the next block these consensus rules changes will take effect:
+
+* Remove topological transaction order constraint, and enforce canonical transaction order
```
zander: The removal of `topological transaction order` is not an improvement.

You want to remove information gathered during the mempool-fill time for practically free. This information is used during validation to make sure we can parallelize the validation of those transactions not in 'topological order'.

When you remove this information then it takes an exponential amount of time to find it again (based on number of transactions inputs+outputs in the block). Which is a huge negative.
A draft (designed on paper) solution to do the inserts first and the deletes later doesn't make any sense in ANY UTXO implementation because the deletes will end up being 100% serialized (due to locking). So similarly there, good bye parallel validation.

A simple solution is to sort the transactions that actually are spent in the same block in a separate, topologically sorted, section. And then sort the rest "canonical".

The Graphene developers have already agreed that this approach would be great for them, giving all the advantages of sorting.

That solution also has the benefit of not being a protocol fork.

Bottom line; **Please don't remove topological transaction order, it breaks known implementations**.

jtoomim: Why do you say that any UTXO implementation will end up with inserts serialized? Concurrent hashtables are a thing; you can set one up with a mutex on each bucket, for example. Lock-free hashtables are also possible using atomic swaps.

zander: jt, that doesn't really address the problem of removing the ordering information Satoshi put in the protocol.

But, concurrent hashtables have no implementation in the standard library or in boost (according to google), and while theoretically people could build a new container for this, that should not be required to make a well-performing implementation for the consensus rules.

jtoomim: It doesn't seem to be important information. The transaction set naturally has the structure of a DAG, not a list. Encoding transactions in blocks as a list adds unnecessary information content to the block encoding. Getting rid of unnecessary information seems like a good idea. Not needing to check that transactions are topologically ordered seems like a win, too -- it's a lot easier to check that transactions are sorted by hash than it is to check that the transaction sorting is a valid topological sorting, especially for parallel algorithms.

By the way, it does not take exponential time to (re)generate a topological sorting of elements. It's actually O(n) versus transaction count for serial algorithms, and O(log^2(n)) versus processor count for parallel implementations. 

We're already using a library outside of boost and std:: to store our UTXOs: LevelDB. It seems within reason to include a concurrent hashtable library. Many such libraries exist in single header form with APIs identical or similar to std::unordered_map, so replacement should be pretty easy. I did that once with sparsepp a few weeks ago, and it wasn't hard. Moreover, std::unordered_map is relatively slow and memory-inefficient, so switching to a more modern hashtable implementation can improve performance even without parallelization.

zander: > It doesn't seem to be important information. 

You should at minimum show the courtesy to read other peoples posts.

> The transaction set naturally has the structure of a DAG, not a list

Yes, as I said there are two types of nodes. One type is the 99% which can be reordered at will since in graph language, they are not related to anyone. The rest of the nodes do, and those need we need to know the ordering for.

>  Getting rid of unnecessary information seems like a good idea. 

Tx 1 spends Tx2, thus you need to process TX1 before Tx2 or you will report that Tx2 spends non-existing funds.  That ordering is really quite important.

> it's a lot easier to check that transactions are sorted by hash than it is to check that the transaction sorting is a valid topological sorting, especially for parallel algorithms.

This misses the fact that no implementation spends even a single cycle today doing this check. So adding a validation rule about sorting **makes things slower**.  Because we don't check sorting today.

Bottom line;

* this rule adds validation for ordering. There is no such validation today.
* this rule removes data needed for parallel validation which makes PV much more costly.
* this rule doesn't help ANY software or technology.

JToomim may have a point that technology could be used to make this not a big issue, but that software should be written, tested and seen to be better than the PV software we have today before a protocol change is made.

Doing a change based on what some developers theorize should be better. without any actual practical experience to back it up,  is really irresponsible. This is not how you deal with a 10 billion dollar industry.

spec/2018-nov-upgrade.md
Mengerian: Yes, wallets that follow the upgrade should not have to change anything.

I will try adding some clarification.

spec/2018-nov-upgrade.md
Mengerian: It definitely should have a "why" somewhere, but is the specification the best place for that?

Or would it be better to have that explanation instead in other supporting documentation, and the bitcoincash.org website?

No problem adding it here though, if that's what everyone thinks is best.

spec/2018-nov-upgrade.md
jtoomim: > Doing a change based on what some developers theorize should be better. without any actual practical experience to back it up, is really irresponsible.

While I may be more optimistic than you about the performance and security of the canonical block idea, I definitely agree with this sentence. No plans should be made to enact a change until it has been shown, with a sample implementation, to be an improvement.

That said, it's acceptable to write a specification for the proposed change in advance of the implementation, which is what this document is supposed to be. We just need to make sure that we are aware that the proposed and specified changes are not yet approved.

spec/2018-nov-upgrade.md
```diff
@@ -0,0 +1,56 @@
+---
+layout: specification
+title: 2018 November 15 Network Upgrade Specification
```
jtoomim: Title should be:

2018 November 15 **Proposed** Network Upgrade Specification

Since the Bitcoin Cash community has not shown overwhelming support for this specification yet, it should be clearly described as a proposal.

spec/2018-nov-upgrade.md
jtoomim: > A simple solution is to sort the transactions that actually are spent in the same block in a separate, topologically sorted, section. And then sort the rest "canonical".

> The Graphene developers have already agreed that this approach would be great for them, giving all the advantages of sorting.

By the way, a while ago I implemented [an encoding scheme for Graphene](https://github.com/jtoomim/orderencode/) which gives approximately the same efficiency on the current CreateNewBlock results that you'd get from the separation of topological and canonical ordering sections. So yes, canonical ordering is not needed for Graphene to work, and probably isn't needed at all.

spec/2018-nov-upgrade.md
ryanxcharles: Great.

"wallets that follow the upgrade should not have to change anything." -> Perfect wording.

spec/2018-nov-upgrade.md
jasonbcox: Motivation is almost always part of a specification.  Links to external supporting docs are always a plus, but a short blurb about the "why" is practically necessary, or it makes evaluating the spec difficult.

Mengerian: OK, I will add that in

spec/2018-nov-upgrade.md
Mengerian: That's a good point.

I don't expect this pull request to get merged for at least a few more weeks, so maybe we should monitor community sentiment, and decide at that point how to characterize it. I'll try to characterize it in a way that seems reasonable.

spec/2018-nov-upgrade.md
```diff
@@ -0,0 +1,56 @@
...
+
+New opcodes OP_CHECKDATASIG and OP_CHECKDATASIGVERIFY will be enabled as specified in [op_checkdatasig.md](op_checkdatasig.md) [2].
+
+## Minimum Transaction Size
+
+Transactions that are smaller than 100 bytes shall be considered invalid.
+
+## Push Only
+
+Transactions shall be considered invalid if an opcode with number greater than 96 (hex encoding 0x60) appears in a scriptSig.
```
dgenr8: If we're serious about this, there are at least 4 known easy malleability vectors that should be closed together:
SCRIPT_VERIFY_NULLDUMMY
SCRIPT_VERIFY_SIGPUSHONLY
SCRIPT_VERIFY_MINIMALDATA
SCRIPT_VERIFY_CLEANSTACK
Reducing malleability improves 0-conf because it makes it harder for miners to invalidate a transaction (knowingly or unknowingly) by confirming a clone of its parent.

deadalnix: +1

spec/2018-nov-upgrade.md
deadalnix: https://blog.vermorel.com/journal/2018/6/12/canonical-transaction-ordering-for-bitcoin.html

spec/2018-nov-upgrade.md
dgenr8: @Mengerian @jasonbcox The vulnerability only affects transactions that are exactly 64 bytes long (because that is also the length of 2 concatenated hashes).  If the solution is to restrict tx size, `size == 64` should be banned, instead of `size < 100`.

A better solution IMHO is to design a more secure merkle proof and migrate to it over time.

jasonbcox: As far as I know, it's not possible to generate transactions smaller than 100 bytes.  There's no reason to allow other malformed transactions, so this limit should be safe.  Also, this limit can be loosened later if such a use-case is discovered in the future.

jasonbcox:
> Is it late for suggesting getting rid of hybrid 0x06, 0x07 P2PKH addresses' prefixes by considering those transactions temporarily nonstandard until making them invalid at some point?

I think so.  The proposals in this spec have been discussed for some time now, so this change would be best proposed for the next hardfork.

spec/2018-nov-upgrade.md
dgenr8: @jasonbcox It is sloppy in the extreme to fix a non-consensus problem with a hard consensus rule.

We learned that loosening a hard limit like this may require a 2 year civil war.  SPV proofs are flawed, so that is what should be fixed.

Like the block size limit, this is broader than any script or validation restriction.  This limit would apply across any conceivable transaction format.  Innovation toward smaller transactions is a direct scaling vector.

spec/2018-nov-upgrade.md
awemany: I like to take the opportunity to point out this very preliminary analysis of one aspect of this change that I did:

https://github.com/awemany/valorder

@deadalnix: Have you done a deeper analysis of this aspect? Other aspects?

I have to admit that it is my fault that I should have been much more sceptical earlier. The fact that the positives of this change are emphasized whereas the negatives (like the above) are dropped under the table tells me that there might be some group think in action. No offense meant.

spec/2018-nov-upgrade.md
ftrader: @Mengerian , could you explain what happened since then, that caused this specification to be published without a 'proposed' tag?

I find myself agreeing with people who are reasonably critical of this, such as to publish on 'bitcoincash.org' without pointing out that the specification for the upcoming upgrade is still contended.

markblundeberg:
A note on the clean stack rule: I believe this will make it impossible to recover accidental spends to segwit P2SH addresses made on BCH network. A segwit [P2SH-P2WPKH](https://bitcoincore.org/en/segwit_wallet_dev/) address has the following two-push redeem script: OP_0 PUSH(pubkeyhash).

spec/2018-nov-upgrade.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Canonical Transaction Order
+
+With the exception of the coinbase transaction, transactions within a block shall be sorted by Transaction ID in ascending order. The coinbase transaction shall be the first transaction in a block.
+
+## OpCodes
+
+New opcodes OP_CHECKDATASIG and OP_CHECKDATASIGVERIFY will be enabled as specified in [op_checkdatasig.md](op_checkdatasig.md) [2].
+
+## Minimum Transaction Size
```
zquestz: Can someone enlighten me on why this is 100 bytes and not 64? All the literature I read about this Merkle tree vulnerability is for 64 byte transactions. Would like to know why this is 100. Thanks in advance.

dgenr8:
@markblundeberg Very good point. Someone will eventually lose a lot of money because of this. You can argue it's their own fault but since more malleability holes still remain it begs the question of why fork in CLEANSTACK right now.

markblundeberg:
> @markblundeberg Very good point. Someone will eventually lose a lot of money because of this. You can argue it's their own fault but since more malleability holes still remain it begs the question of why fork in CLEANSTACK right now.

Just to follow up on this --- it looks like the person who will stand to lose the most money from this is [the sneaky miner who snatched up this poor guy's BCH, along with many other peoples' BCH](https://www.reddit.com/r/btc/comments/9diobf/raw_bitcoin_cash_segwit_transaction/), in [this block(https://bch.btc.com/000000000000000001a13b7ffba2447d7b55132117f9cccd950f54ae041272a7). Apparently there is no honest miner offering a segwit recovery anymore. Starting from November that miner will have to give up on snatching these coins.

dgenr8:
> Apparently there is no honest miner offering a segwit recovery anymore. 
I noticed @jtoomim was [willing to help](https://www.reddit.com/r/btc/comments/9diobf/raw_bitcoin_cash_segwit_transaction/e5ij2sk/).

spec/2018-nov-upgrade.md
sickpig: Guess @deadalnix, @jasonbcox or @schancel could chime in. We were in agreement to fix it but we had different idea on the details, then ABC in the process of respecting releases timeline included the 100 byte limit in the code.

As a matter of fact we could have put a different constraint, i.e. `!= 64` or used [Sergio Demian Lerner's proposal](https://bitslog.wordpress.com/2018/06/09/leaf-node-weakness-in-bitcoin-merkle-tree-design/) to fix it w/o the need of a fork.

markblundeberg: I am thinking of an application using BCH that would make use of very short anyone-can-spend redeem scripts as 'notification addresses' for lite clients. For example the redeem script would be PUSH(8 bytes) or even PUSH(4 bytes), spendable with empty scriptsig. The idea with these addresses is that anyone (miners included) can grab the dust.

Ideally miners could spend these utxos using one-input, zero-output transactions that just add them to the collected fees. Such transactions would be only 61 or 57 bytes long. So, I think the `!= 64` constraint would make most sense.

awemany: @markblundeberg: I have stumbled over the 100tx limit while doing my "Zero Conf Forfeit" transaction as well. @schancel told me it should be easy to remove again (and technically, I agree).

I also think we should definitely allow small transactions in the future, even if there are plans to reserve space in the merkle / to-be-merklix tree.

A P2PKH -> P2SH transaction is impacted by the 100 byte limit.

As I compromise, I'd like to put onto the agenda to remove this limit again when Sergio Demian Lerner's proposal is implemented (or similar) and make this the goal for the next HF.

SubbieMasse: Thank

markblundeberg: There have been 29000 transactions under 64 bytes so far:
https://blockchair.com/bitcoin-cash/transactions?q=size(..63)#

Only five have been 64 bytes exactly:
https://blockchair.com/bitcoin-cash/transactions?q=size(64)#

Proposal for next May fork: fix the size limit so it is `!= 64`.

markblundeberg:
> Is it late for suggesting getting rid of hybrid 0x06, 0x07 P2PKH addresses' prefixes by considering those transactions temporarily nonstandard until making them invalid at some point?

I remembered this question of yours from long ago, @DesWurstes . :-) Actually, it turns out that hybrid pubkeys were even already banned at consensus layer from STRICTENC rule, which was adopted in the Aug 2017 hardfork. I only realized this recently when digging through the ABC code.

DesWurstes:
Thanks for letting me know, I still hadn't noticed before I saw your comment. 👍

PR #103, opened 2018-08-20
0.18.0 link to abc

maplesyrupsucker:


Mengerian:
Could you move the "Compatible Implementations" to after the Spec links?

Also, I think it would be better if "Compatible Implementations" were not a link, just "Bitcoin ABC 0.18.0"

PR #130, opened 2018-10-03
added memopay.xyz and SLP as services

tracyspacy:


BUYMERCIER:
Please submit only one complete PR and named properly.

PR #132, opened 2018-10-06
Hackathon promotion modal

BUYMERCIER:


css/custom.css
```diff
@@ -75,6 +75,19 @@ p.release {
   background-attachment: fixed;
 }
 
+.close-fix {
+  position: relative;
+  top: -15px;
```
BUYMERCIER: Dirty, but temporary

PR #144, opened 2018-10-17
Spec for re-enabled opcodes, May 2019 upgrade

jasonbcox:
This is a markdown version of nChain's spec for re-enabling opcodes which was proposed past the deadline for 2018 November protocol changes.  Adding it here for consideration and review for the May 2019 upgrade.

### DIFF
--- a/spec/may-2019-reenabled-opcodes.md
+++ b/spec/may-2019-reenabled-opcodes.md
@@ -1,0 +1,210 @@
+---
+layout: specification
+title: Specification for Re-enabling old Opcodes, May 2019
+date: 2018-08-03
+version: 1.0
+updated: 2018-08-13
+---
+
+# Introduction
+
+In May 2018 several disabled opcodes were reintroduced to the Bitcoin Cash scripting engine2. The scope of that change was limited in order to focus developer attention rather than attempting to reintroduce all of the disabled opcodes at once. This specification expands upon that change by reintroducing additional opcodes.
+
+This specifications describes the opcodes that are proposed to be added in the May 2019 protocol upgrade.
+
+The opcodes to be added are:
+| Word | OpCode | Hex | Input | Output | Description |
+| :--- | :----- | :-- | :---- | :----- | :---------- |
+| OP_MUL | 149 | 0x95 | n1 | n2 | out | Multiplies two numbers |
+| OP_RSHIFT | 153 | 0x99 | b | n | out | Right shift b by n bits |
+| OP_LSHIFT | 152 | 0x98 | b | n | out | Left shift b by n bits |
+| OP_INVERT | 131 | 0x83 | b | out | Bitwise | NOT |
+
+## Data types
+
+Script data values are byte sequences but may be interpreted as numeric values by some opcodes. Specification of the data types used by Script is beyond the scope of this document.
+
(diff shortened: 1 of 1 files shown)


spec/may-2019-reenabled-opcodes.md
```diff
@@ -0,0 +1,210 @@
...
+Description: Right shift b by n bits.
+
+```
+b n OP_RSHIFT → out,
+where out is the byte sequence created by a right shift of the binary bit pattern b by n bits.
+```
+
+There is no limit imposed on the value of n except that n must be a valid numerical value and must be non-negative. If n is negative, the script fails.
+
+If n is greater than the number of bits in the byte sequence, the result is a zero filled byte sequence equal in length to the original byte sequence. This is consistent with the interpretation of b n OP_RSHIFT as equal to b 1 OP_RSHIFT repeated n times.
```
jasonbcox: I think this is lightly implied by the spec here, but @jtoomim made a good point to ensure that logical bit shift was specified rather than arithmetic shift.  This should apply for both OP_RSHIFT and OP_LSHIFT.

spec/may-2019-reenabled-opcodes.md
```diff
@@ -0,0 +1,210 @@
...
+updated: 2018-08-13
+---
+
+# Introduction
+
+In May 2018 several disabled opcodes were reintroduced to the Bitcoin Cash scripting engine2. The scope of that change was limited in order to focus developer attention rather than attempting to reintroduce all of the disabled opcodes at once. This specification expands upon that change by reintroducing additional opcodes.
+
+This specifications describes the opcodes that are proposed to be added in the May 2019 protocol upgrade.
+
+The opcodes to be added are:
```
Mengerian: Need a newline here so that the table renders properly
```suggestion
The opcodes to be added are:

```

spec/may-2019-reenabled-opcodes.md
```diff
@@ -0,0 +1,210 @@
...
+* The number of elements on stack is unchanged.
+
+### Unit Tests
+
+1. {} OP_INVERT → {}
+2. b OP_INVERT → bitwise invert(x) for various b and various sizes of b. E.g. {0x80, 0x12, 0x34} → {0x7F, 0xED, 0xCB}
+
+# References
+
+[1] [https://en.bitcoin.it/wiki/Script#Opcodes](https://en.bitcoin.it/wiki/Script#Opcodes)
```
Mengerian: Also blank line here so that the references render on separate lines
```suggestion
[1] [https://en.bitcoin.it/wiki/Script#Opcodes](https://en.bitcoin.it/wiki/Script#Opcodes)

```

spec/may-2019-reenabled-opcodes.md
```diff
@@ -0,0 +1,210 @@
+---
+layout: specification
+title: Specification for Re-enabling old Opcodes, May 2019
+date: 2018-08-03
```
schancel: Please put the activation date at the beginning for lexicographical sorting.  I intend to fix the other specifications.

`20190515-reenabled-opcodes.md`

Mengerian: How about putting dashes between the numbers:

`2019-05-15-reenabled-opcodes.md`

Makes it more human-readable as a date, and still sorts properly

schancel:
I'm stealing this pull request.

spec/may-2019-reenabled-opcodes.md
schancel: And harder to type :/

schancel:
I've incorporated existing feedback

PR #145, opened 2018-10-24
:sparkles: Add the draft of Deep Link Payment Protocol spec

usatie:
Propose a specification for Deep Link Payment Protocol.

This protocol is meant to be used for communication between mobile wallet app and other applications.

With this protocol, the communication wallet application and other applications will be very smooth.

This PR is meant to start discussion.

demo

### DIFF
--- a/spec/deep-link-payment-protocol.md
+++ b/spec/deep-link-payment-protocol.md
@@ -1,0 +1,151 @@
+```
+BIP: ?
+Layer: Applications
+Title: Deeplink Payment Protocol
+Author: Shun Usami<usatie@yenom.tech>,
+        Taiki Uchida<yuiki@yenom.tech>,
+        Aoi Serikawa<seri@yenom.tech>
+Comments-URI: https://hackmd.io/s/BJObJntjQ
+Status: Draft
+Type: Standards Track
+Created: 2018-10-21
+```
+
+## Abstract
+
+Basically this is just a [BIP21(URI scheme)](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki) like deeplink scheme to wallet applications with a callback url.
+When the wallet app completed the requested behaviour, the wallet app ping the callback URL.
+![demo](https://i.imgur.com/yXBsfJL.gif)
+
+## Motivation
+Currently, the payment experience on mobile devices are very bad, because users have to take many steps like this.
+- 1. Copy the address.
+- 2. Open a wallet app.
+- 3. Paste the address.
+- 4. Type the amount to send.
+- 5. Send to the address.
+- 6. Re-open the source application.
+
+This protocol dramatically reduces the user's required steps.
+- 1. Opens the deep link to a wallet app.
+- 2. Confirm payment and automatically returns to the source app.
+
+## Specification
+
+### Deep link Scheme
+This deep link scheme is protocol for communication between wallet application and another application.
+
+i.e.
+`bch-payment:{parameters}`
+`bch-sign:{parameters}`
+
+
+| title | value |
+|:---:|:---:|
+|Custom URL scheme|bch-{type}|
+
+
+| type | explanation |
+|:---:|:---:|
+| payment | request a payment to wallet application |
+| sign | request signing of message to wallet application |
+| coming soon | coming soon ...|
+
+
+### Mobile App to Wallet App Communication
+![abstract:mobile-application](https://i.imgur.com/SPywIQE.png)
+
+#### 1. Source app generates deep link to wallet app
+The source app generates a deep link to wallet apps.
+i.e.
+`bch-payment:{addr}?amount={amount}&callback={callback_url}`
+
+`addr` is a cashaddr without scheme.
+i.e.
+`qznj9wsazh379qa3dkph7qvtxdx5emkrxvsxhnyall`
+
+|Query key| type | memo | example value|
+|:---:|:---:|:---:|:---:|
+|amount| Decimal | Coin value to send | 0.0001 |
+|callback| String | deep link | yenom://payment-result |
+
+#### 2. Wallet app make the payment
+The wallet app build the payment transaction and broadcast it.
+
+
+#### 3. Wallet app opens the deep link to the source app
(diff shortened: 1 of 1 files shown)


usatie:
These are useful referencing materials to start discussion.

[BIP70](https://github.com/bitcoin/bips/blob/master/bip-0070.mediawiki)
[bitpay/jsonPaymentProtocol](https://github.com/bitpay/jsonPaymentProtocol/blob/master/specification.md)
[unwriter/fatURI](https://github.com/unwriter/fatURI)


Things to be considered are these
- How to prevent MITM(Man In The Middle) attack?
- passing feePerByte info?
- passing minConf info?
- passing outputs?
- passing rawtx?

usatie:
We already got some pieces of advice.

- Better to specify the encoding format for URL
- How to deal with the case when txid is malformed?
- Better to pass session id hash instead of session id itself. Or maybe more general argument name like  "nonce" or "merchant_data" is better.

usatie:
I think the level of complexity and flexibility is tradeoff, so we should set different types for different purposes.

i.e.
`bch-payment://` type contains only amount and address.
`bch-outputs://` type contains several hex format of outputs.
`bch-rawtx://` type contains only hex format of rawtx.

spec/deep-link-payment-protocol.md
```diff
@@ -0,0 +1,151 @@
...
+Type: Standards Track
+Created: 2018-10-21
+```
+
+## Abstract
+
+Basically this is just a [BIP21(URI scheme)](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki) like deeplink scheme to wallet applications with a callback url.
+When the wallet app completed the requested behaviour, the wallet app ping the callback URL.
+![demo](https://i.imgur.com/yXBsfJL.gif)
+
```
jasonbcox: I recommend adding a background section to give a quick explanation of what deeplinking is.  https://en.wikipedia.org/wiki/Mobile_deep_linking is a good place to start.

jasonbcox:
Regarding your question on how to prevent MITM attacks: This is not necessary as it is done for you.  Both iOS and Android operating systems have an app selection process (usually involving a prompt to the user) that determines where the deeplink request is forwarded to.  Assuming the OS is not compromised, other apps on the device are unable to modify the request in transit.

spec/deep-link-payment-protocol.md
```diff
@@ -0,0 +1,151 @@
...
+#### 4. Source app verify the tx with txid
+Source app should verify the payment is done.
+
+
+### Web App to Wallet App Communication
+![abstract:web-application](https://i.imgur.com/tYHey1y.png)
+
+#### 1. Source web site generates deep link to wallet app
+
+`bch-payment:{addr}?amount={amount}&webhook={webhook_url}&callback={callback_url}`
```
jasonbcox: Any reason the {addr} is not part of the query param set? Is that because it is required and the query params are optional?

spec/deep-link-payment-protocol.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Rationale
+This protocol is quite simple.
+Alternative [BIP70](https://github.com/bitcoin/bips/blob/master/bip-0070.mediawiki#Motivation) like design can be considered.
+
+## Forward compatibility
+Adding optional parameters can be done by simply adding them.
+Adding required parameters or destructive change can be done by simply changing the type name to new one.
+i.e.
+`bch-payment` -> `bch-payment-v2`
```
jasonbcox: Similar to the bch-{type} cons above, a new payment version would be unrecognized by old wallets.  I recommend making 'version' a query param, with default value of '1'.  This way, apps can error out with an informative message such as "New version of bitcoin-cash payment is not supported! Please contact the app developer."

usatie:
@gitcash send 1000 bits to @jasonbcox 

I think MITM attack can be done by browsers and browser extensions.
What do you think?

> Regarding your question on how to prevent MITM attacks: This is not necessary as it is done for you. Both iOS and Android operating systems have an app selection process (usually involving a prompt to the user) that determines where the deeplink request is forwarded to. Assuming the OS is not compromised, other apps on the device are unable to modify the request in transit.

GitCash:
Hey **jasonbcox**, user **usatie** tipped you **1000 bits** in Bitcoin Cash ( ~ **$0.439** ).

*[Click here to claim it!](https://gitcash.io/claim/jasonbcox)*

You can also add the "thumbs down" reaction to **usatie**'s comment above to prevent future tips.

spec/deep-link-payment-protocol.md
```diff
@@ -0,0 +1,151 @@
...
+This deep link scheme is protocol for communication between wallet application and another application.
+
+i.e.
+`bch-payment:{parameters}`
+`bch-sign:{parameters}`
+
+
+| title | value |
+|:---:|:---:|
+|Custom URL scheme|bch-{type}|
```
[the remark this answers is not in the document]
usatie: The problem is iOS. iOS users cannot choose application to open a deep link when multiple apps are supporting the link. In such case, iOS kind of randomly choose the app to open the link.

When users have multiple wallet apps, and some of them are supporting the {type} but some of them are not supporting the {type}. Since all the wallet seems to be supporting the link.
So users want to know if a wallet app supports {type} before opening the link.

spec/deep-link-payment-protocol.md
usatie: There is no reason about it. Let me fix it.

spec/deep-link-payment-protocol.md
usatie: This is intended to be so.
As I stated above, the problem is iOS.
>a new payment version would be unrecognized by old wallets.

MissMonacoin: How about BIP21-style prefix?
If req- prefix was found and the old app wasn't able to parse the field name, the app can mark as invalid.
Many bitcoin cash wallet has already supported BIP21 so I think it would be better to follow BIP21.
https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki#forward-compatibility

PR #250, opened 2019-01-06
added cashslide

kosinusbch:
lets you slide into them on-chain dms (to the protocol list, since its moved to this repo for whatever reason)


kosinusbch:
Don't know who to @, but changes to this specific file shouldn't really be something that requires weeks of review. @maplesyrupsucker

maplesyrupsucker:
@Mengerian care to review?

Mengerian:
Hmmm, I'm not sure how to review this. There's doesn't seem to be any information about what it is, and the url's don't lead anywhere.

We'll need to think of how to deal with this if there are many requests for the Lokad ID numbers.

kosinusbch:
@Mengerian It's just a list of reserved protocols. There is nothing to review, it was previously hosted in another repo then was merged with this for whatever reason

Mengerian:
OK, that's fine. But should there at least be some link to what the protocol actually is? The url "bch.sx" doesn't seem to lead anywhere. Is there any information you can link to provide some idea about Cashslide is?

I'll approve it for now, but this process seems open to potential abuse.

kosinusbch:
It's a on-chain encrypted messaging system. bch.sx is a clearweb proxy for bch-dns domains, which is another protocol I've reserved for on-chain domain and file hosting. Protocol obviously being reserved before the site or spec is up.

Mengerian:
Thanks for the explanation @kosinusbch 
That makes sense, nice to know what it is for!
All the best with the project, sounds interesting.

PR #259, opened 2019-01-16
Generalized CashAddr

m4ktub:
The goal of this pull request is to generalize the CashAddr specification so that it can be used with new address types, like stealth/reusable addresses that need more than an hash as payload. The changes can be summed up to:

* Expand the purpose of the prefix to refer to use case differentiation (e.g.: SLP).
* Change the interpretation of the 3 least significant bits of the version byte, now renamed type byte, to allow a different meaning for each type.

Although the goal was to preserve backward compatibility, the revision of the size bits mean the test implied by some test vectors (those with type 15) changed, that is, addresses with an unknown type can no longer be length checked from the size bits alone. For that reason those test vectors were segregated to a new section and with new payloads.

### DIFF
--- a/spec/cashaddr.md
+++ b/spec/cashaddr.md
@@ -4,33 +4,41 @@ title: Address format for Bitcoin Cash
 category: spec
 date: 2017-10-13
 activation: 1515888000
-version: 1.0
+version: 1.1
 ---
 
 ## Abstract
 
-This document describes a proposed address format to be used on Bitcoin Cash. It is a base32 encoded format using BCH[[1]](#bch) codes as checksum and that can be used directly in links or QR codes.
+This document describes CashAddr, a proposed address format to be used on Bitcoin Cash. It is a base 32 encoded format using BCH [[1]](#bch) codes as checksum and that can be used directly in links or QR codes.
 
-This format reuses the work done for Bech32[[2]](#bip173) and is similar in some aspects, but improves on others.
+This format reuses the work done for Bech32 [[2]](#bip173) and is similar in some aspects, but improves on others.
 
 ## Specification
 
 The address is composed of
-1. A prefix indicating the network on which this address is valid.
+1. A prefix indicating the purpose or network of the address for compatibility reasons.
 2. A separator, always `:`
-3. A base32 encoded payload indicating the destination of the address and containing a checksum.
+3. A base 32 encoded payload with the destination of the address and a checksum.
 
 ### Prefix
 
-The prefix indicates the network on which this addess is valid. It is set to `bitcoincash` for Bitcoin Cash main net, `bchtest` for bitcoin cash testnet and `bchreg` for bitcoin cash regtest.
+The prefix indicates the purpose or network of the address. It guides users into choosing an application and allows applications to verify support. The following prefixes are reserved for the specified address types in the three Bitcoin Cash networks:
+* `bitcoincash` for mainnet, 
+* `bchtest` for testnet, and 
+* `bchreg` for regtest.
 
 The prefix is followed by the separator `:`.
 
 When presented to users, the prefix may be omitted as it is part of the checksum computation. The checksum ensures that addresses on different networks will remain incompatible, even in the absence of an explicit prefix.
 
-### Payload
+### <a name="payload"></a> Payload
 
-The payload is a base32 encoded stream of data.
+The payload is a base 32 encoded stream of data composed of 3 elements:
+1. A byte indicating the type of address.
+2. A variable number of bytes with address data.
+3. A 40 bits checksum.
+
+The following table is used for encoding.
 
 |     | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 |
 | --: | - | - | - | - | - | - | - | - |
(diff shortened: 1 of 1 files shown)


markblundeberg:
CashAddr is specifically concerned with actual addresses, i.e., it's a human representation of certain standard scriptPubKey forms. My suggestion is that if the prefix is not bitcoincash: / bchtest: / etc., then there should be no restrictions at all, and it is outside the scope of CashAddr so this document does not need any changes.

If people want to come up with an "address" scheme for some other system that is not actual addresses, nothing stops them from using a URI prefix `mystealthaddr:` with a base32 encoding. What comes after that is up to their specification (they can of course borrow concepts like the checksum, if they want). But just because something is URI prefixed with base32 payload, doesn't mean it's CashAddr.

Besides that it seems strange to preemptively restrict the first bit like this for all protocols. For example, maybe I want my stealth "address" system to look like `stealthaddr:s....`. The spec forbids this, so why should I follow this spec?

m4ktub:
> CashAddr is specifically concerned with actual addresses, i.e., it's a human representation of certain standard scriptPubKey forms.

I understood that perspective better from my interactions with Deadalnix over Twitter.

> My suggestion is that if the prefix is not bitcoincash: / bchtest: / etc., then there should be no restrictions at all,

That was also my suggestion, in the end. Either people agree with this change or the spec should be reviewed to remove the specification of the first byte in the payload. I have doubts about the ability to make a v1.1 in that case but I haven't investigated the implication with more detail yet.

> and it is outside the scope of CashAddr so this document does not need any changes.

I disagree. This specification makes `simpleledger` addresses a perfectly valid CashAddr. Those are even accepted by Copay's library which implements this spec.

> If people want to [...]

I believe that having a specification that allows legitimate reuse for several purposes without ambiguity is a plus for adoption and usability. Currently, as I see it, this specification only states its scope through the inability of allowing anything else while creating the expectation of being the "one" format for everything the user perceives as addresses (because a random string of characters, or QR code, is indistinguishable from another to users).

Going back to the first point, I'll try to review the specification again from that perspective. It seems harder to make a backward compatible and coherent specification but I haven't tried yet.

BigBlockIfTrue:
I think it makes a lot of sense to integrate stuff like stealth addresses into the CashAddr spec by defining new address types beyond P2PKH and P2SH:

* **User experience for the address field.** While stealth addresses do not appear as such on the blockchain, from an UX point of view, they still function as addresses. It is something a user will have to put in an address field. It should therefore look like an address. The similarity to any other Bitcoin Cash address should be maximal. This can only be achieved if they're defined by the same CashAddr specification in a mutually consistent way.
* **Optionality of the prefix.** On the mainnet, the `bitcoincash:` prefix is optional, avoiding unnecessary verbosity if it's clear from the context that something is a Bitcoin Cash address. Having a mandatory prefix for some address types is unnecessarily confusing. Just imagine P2PKH and P2SH would have different prefixes and P2SH's would be mandatory; it would probably be rather weird the first time you use P2SH, non-technical users who didn't use the optional prefix before might even think `bitcoincash-p2sh:` were some kind of hacking code. I am not aware of any traditional payment system where account numbers have a mandatory prefix like this either.
* **Clean URI scheme handler association.** If every address type requires a separate handler registration mapping the URI scheme to the wallet software, wallet software will need to register every one of them separately, which is both error-prone. It is furthermore collision-prone with other IANA registrations or other software, and the likelihood of such collisions can only be reduced by making the prefixes long, which in turn damages usability. Unlike SLP, the different address formats are intended to spend the same Bitcoin Cash, so it makes sense to use it as a single URI scheme. If the associated wallet software does not support the address type, it can simply notify the user.

deadalnix:
Ok, First thing first, what is actually needed to encode a stealth address ?

markblundeberg:
@deadalnix At minimum a stealth payment code system could be easily done with encoding a single pubkey (33 bytes) -- everything could be based on / derived from that. BIP47 payment codes are however [80 bytes long](https://github.com/bitcoin/bips/blob/master/bip-0047.mediawiki#version-1) because BIP47 likes to make things overcomplicated.

m4ktub:
There's no formalized specification. Looking at [Copay's PR#1893](https://github.com/bitpay/bitcore/pull/1893), which uses [Peter Todds' proposal](https://github.com/bitpay/bitcore/blob/bd943d8f97efc353c6415413b7c65152dd2cee8c/packages/bitcore-stealth/bip-stealth.md), we can see that the payload would contain:

* Options (1 byte)
* Scan Public Key (33 bytes)
* Number of Spend Public Keys [Nk] (1 byte)
* _Nk_ x Spend Public Keys (33 bytes)
* Number of Required Signatures (1 byte)
* Prefix Length In Bits [Pb] (1 byte)
* Prefix Data (Pb / 8 bytes)

This format is intended to allow P2PKH, multi-sig P2SH, and SPV filtering through an output prefix. The minimum payload size, for this format, is around 71 bytes. There's a size above which it's impractical to encode the stealth address. I think @matiu suggested considering uses cases of n-of-m up to m = 3.

m4ktub:
@deadalnix @markblundeberg With PR #266 there's now the revisions under two different approaches. 

This one tries to keep the payload structure, for all prefixes, but redefines the type byte to have the ability to create new types. The one in PR #266 unconstraints the payload structure for prefixes not covered by the specification.

deadalnix:
Looking at the spec from @matiu it really does sounds like the first step here is to actually figure out what we want for stealth addresses to begin with. Depending on the result, we can see what kind of addresses is the best. It does look like that what will be sent will very much be some kind of complex datastructure, especially if we want to support multisig and alike. Maybe the best is just to send a JSON over the wire ?

PR #266, opened 2019-01-30
Open CashAddr

m4ktub:
This pull request provides a revision of CashAddr that intends to make the specification less ambiguous while, at the same time, allowing CashAddr to be used as a general form of address encoding on Bitcoin Cash. The changes can be summed up as:

* Reserve three prefixes for the address types also defined in the specification.
* Unconstrain the payload structure for any other prefix.

This revision makes previously invalid payloads now valid (eg. `prefix:x64nx6hz`). It also makes the specification of `simpleleger` addresses slightly more ambiguous, although arguably still valid. Like in the PR #259, due to the unconstraint of the payload for other prefixes, some test vectors were changed as they can no longer be type and length checked.

After a brief review of some libraries, although theoretically less backward compatible it seems this revision is, in practice, as backwards compatible as the one in PR #259. Most libraries are limited to the 3 Bitcoin Cash networks and their corresponding address prefixes.

In this specification new address schemes, like stealth addresses, could look like this: `stealth:9vqq80nex6callf0srmgrsmfvxmkd30354s0ukmft5c2hjhtdafth7weqypyrttwfj3es6xufs3fqdwlz4lqnf9wr6ldwv38fdnph4rhu34tclgpqqqqfrvvj9ll`.

### DIFF
--- a/spec/cashaddr.md
+++ b/spec/cashaddr.md
@@ -38,11 +38,13 @@ The payload is a base 32 encoded stream of data composed of 2 elements:
 2. A 40 bits checksum.
 
 <pre>
- data             checksum
-+------ ~ -------+---------+
-|    N bytes     | 5 bytes |
-+------ ~ -------+---------+
+ data             padding  checksum
++------ ~ -------+--------+---------+
+|    N bytes     | P bits | 40 bits |
++------ ~ -------+--------+---------+
 </pre>
+
+Since the data is base 32 encoded, the last group of 5 bits may not be complete. In that case a padding of zeros is used. For example, if 2 bytes are encoded, 4 bits of padding are added to complete the last group and the data part would be represented by 3 characters. The last character would necessarily be either _q_ or _s_.
 
 The following table is used for the encoding.
 
--- a/spec/cashaddr.md
+++ b/spec/cashaddr.md
@@ -108,7 +108,7 @@ The following addresses can be used as test vectors for checksum computation sin
 
 ## <a name="addresstypes"></a>Address types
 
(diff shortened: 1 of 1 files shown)


markblundeberg:
There is a slight conceptual error in this one. You write: 
```
 data             checksum
+------ ~ -------+---------+
|    N bytes     | 5 bytes |
+------ ~ -------+---------+
```

However actually it's structured like this:
```
 data             checksum
+------ ~ -------+---------+
|    M chars      | 8 chars |
+------ ~ -------+---------+
```

In other words, the checksum is appended *after* conversion to base32.

The distinction is important in the case where the number of bits preceding checksum (8N) is not a multiple of 5. For example when we have normal P2PKH addresses (N = 21, so 168 bits) that leaves only 3 bits left over on the last base32 character of the payload. So if you look at the 9th-from-last character on such an address you'll see it only ever takes on one of 8 values: q, p, z, r, y, 9, x, 8, and not any of the other 24 values.

m4ktub:
You're right about my error, the payload is not necessarily byte aligned (8N = 5M) and I assumed it was.

In that part I'm referring to the data, and not the representation, so I can also add the variable padding to the diagram as bellow. I believe it's important to make explicit that the encoded data is actually a series of bytes since, in general, there would no way to figure "P".

<pre>
 data             padding  checksum
+------ ~ -------+--------+---------+
|    N bytes     | P bits | 40 bits |
+------ ~ -------+--------+---------+
</pre>

Regarding your example, just a note that, the 8 possible characters are actually q, y, g, v, s, 5, c, and u.

markblundeberg:
Aha! You're right, good catch about those characters :-D. And yes, I think that new representation is clearer. I have to read over the doc again, though...

PR #268, opened 2019-02-02
[spec] Rename May 2018 Upgrade files

Mengerian:
Adding hyphens makes the date more human-readable, and still sorts properly.


maplesyrupsucker:
Love human readable dates. 👍

PR #275, opened 2019-02-10
Roadmap PNG > html/css

maplesyrupsucker:
Converted the graphic from PNG to a more easy to edit chunk of markup and styles. Will allow us to use translated strings and gracefully handle roadmap localization without the need for edits to adobe xd assets.

### DIFF
--- a/_includes/roadmap/roadmap.html
+++ b/_includes/roadmap/roadmap.html
@@ -24,10 +24,6 @@
                 </p>
 
                 <p>To become a solid base for application development and innovation, Bitcoin Cash must continuously improve and compete. Working together, we can build a technical foundation to empower Bitcoin Cash to be the best money the world has ever seen.</p>
-
-
-                
-                <img src="/img/bitcoin-cash-protocol-roadmap.png" alt="Bitcoin Cash Roadmap"/>
             </div>
             
         </div>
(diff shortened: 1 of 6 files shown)


_includes/roadmap/roadmap.html
```diff
@@ -24,12 +24,155 @@ <h2>{% t 'headings.roadmap' %}</h2>
...
+                <img src="media-kit/3-bitcoin-cash-logo-ot-medium.png" alt="Bitcoin Cash roadmap">
+                <h3 class="rm-col-title">
+                    Roadmap
+                </h3>
+            </div>
+        </div>
+        <div class="row">
+            <div class="col-sm-12 col-md-4 to-animate rm-scaling">
+                <h3 class="rm-col-title">
+                    Scaling
```
BUYMERCIER: Maybe we could fetch the content from the i18n YAML files so it gets automatically translated here?

maplesyrupsucker: Yea that's the idea - i think we should put that in a separate PR tho so we can just get the updates @Mengerian requested in asap.

We'll need to take the strings from transifex and merge them into the .yml files. 

If you have time tho to tackle that within this PR maybe we can hold off a day or two. Thoughts?

_includes/roadmap/roadmap.html
```diff
@@ -24,12 +24,155 @@ <h2>{% t 'headings.roadmap' %}</h2>
...
+                        <p>(blockchain pruning)</p>
+                        <ion-icon class="underway" name="cog"></ion-icon>
+                    </li>
+                    <li>
+                        <h4>Schnorr Signatures</h4>
+                        <p>(batched signature validation)</p>
+                        <ion-icon class="underway" name="cog"></ion-icon>
+                    </li>
+                    <li>
+                        <h4>Adjustable Block Cap ABC</h4>
```
Mengerian: This should read "Adaptive Block Size"

maplesyrupsucker: Thanks - latest commit resolves this

_includes/roadmap/roadmap.html
BUYMERCIER: I'm on it

PR #310, opened 2019-03-31
Added string to i18n source

merc1er:


merc1er:
Screenshot 2019-03-31 at 11 16 36

Those two strings (roadmap intro and button) were directly in the html document

PR #311, opened 2019-04-02
Create client side filtering spec

cpacia:
This is the spec for client side filtering as implemented in bchd. I'm submitting the PR here so that anyone looking to implement the spec in another client has a spec to base it on. Also, this spec isn't necessarily final and if there is feedback we are open to including it in bchd. 

Specifically, it's unclear if the OP_RETURN additions are useful enough to justify keeping as part of this spec or not. Especially considering that they increase the filter size a little.


markblundeberg:
Is the resulting filter expected to be perfectly defined (i.e., every node gives same filter) ? If so then ExtractDataElements needs to be precisely defined taking into account that scripts may contain: weird push opcodes, non-push opcodes, or unparseable scripts that end mid-push ([example](https://github.com/Electron-Cash/Electron-Cash/issues/829)).

cpacia:
> Is the resulting filter expected to be perfectly defined (i.e., every node gives same filter) ?

Yeah the expectation is every node calculates the filter the same because the lite clients will download both if they get two different ones from different nodes and ban the one that gave them the filter calculated differently from their code.

cpacia:
Here's the function in bchd:
```go
func ExtractDataElements(script []byte) ([][]byte, error) {
	var dataElements [][]byte
	pops, err := parseScript(script) // This errors if any part of the script is invalid. This includes invalid opcodes or pushdatas with an incorrect number of bytes. 
	if err != nil {
		return nil, err
	}
	for _, pop := range pops {
		// All opcodes up to OP_16 are data push instructions.
		if pop.opcode.value <= OP_16 {
			dataElements = append(dataElements, pop.data)
		}
	}
	return dataElements, nil
}
```

markblundeberg:
Looks like that relies on what pop.data is for the special push opcodes -- for example, with OP_1. I suspect it's an empty array if it's like the other script parses I know (which is not the value that gets pushed on stack -- length 1 byte array [0x01]). ...

cpacia:
Yeah I'm looking at it. When I wrote the code I wasn't considering OP_0 - OP_16. Only OP_DATA_1 - OP_PUSHDATA4. And it looks like that's what I ended up with though the code could stand to be more explict. 

Do you see any need to add OP_0 - OP_16 to the filter? Doesn't seem like any compelling use cases. 

Also, the script parsing only fails if the pushdata opcodes have an invalid number of bytes in some way. And a parsing error causes the loop to continue without adding anything rather than adding the data elements that parsed correctly.

markblundeberg:
> Do you see any need to add OP_0 - OP_16 to the filter? Doesn't seem like any compelling use cases.

Indeed, I don't see any value there for flagging their inclusion. (BTW fun fact: OP_RESERVED is standard to use after OP_RETURN ... since it falls between OP_1 and OP_1NEGATE)

>Also, the script parsing only fails if the pushdata opcodes have an invalid number of bytes in some way. And a parsing error causes the loop to continue without adding anything rather than adding the data elements that parsed correctly.

That sounds good. Those are nonstandard anyway btw, so they deserve punishment. :-D

PR #327, opened 2019-04-15
add cashaccounts to protocols.csv

kosinusbch:
There has been an open issue for this for half a year, this should be maintained more to not discourage people from registering their protocol prefixes, which pretty much no one has done since it was taken over by you guys.


Mengerian:
Yes, this should be added in.

PR #328, opened 2019-04-16
Added licho's last will to services

maplesyrupsucker:


Mengerian:
Hmm, I think Licho's Last Will is a really cool project, but maybe it's a bit too experimental for the site's target audience?

KarolTrzeszczkowski:
> Hmm, I think Licho's Last Will is a really cool project, but maybe it's a bit too experimental for the site's target audience?

It is new, but really simple and easy to use. I believe the most difficult step in the usage is the plugin zip file signature verification against my public key, but we expect all Electron Cash users to do it during EC installation. The rest works out-of-the-box. The minimum knowledge level to use it is the understanding of cold storage. Operating Last Will contract is most of the time one click. I really tried to do it as user-friendly as I could. It is new so the errors might still be hiding somewhere, but given the simplicity of it, I doubt anything really bad can happen. It's below 1000 lines of the code. 

If there are additional conditions I need to fulfill, please tell me. I can come back with a landing page, if GitHub is too scary for a regular user, add info to README, etc.

Mengerian:
Thanks for the additional information @KarolTrzeszczkowski that's sounds like it's stable enough to include.

I have asked for a second opinion from @markblundeberg since I think he has looked at your product a bit also.

> If there are additional conditions I need to fulfill, please tell me. I can come back with a landing page, if GitHub is too scary for a regular user, add info to README, etc.

It's not really a requirement for adding it here, but some sort of landing page with a general introduction to how the Last Will plugin works sounds like a good idea. It would probably help you with adoption, and give people something to point new users to.

PR #330, opened 2019-04-17
Update protocols.csv

monsterbitar:
Add donation address for cash accounts.


monsterbitar:
I've added the comma at the end, I think that should do it.

PR #340, opened 2019-05-08
Remove Bitcoin XT

Mengerian:
Bitcoin XT developers have announced that they will not be implementing the upgrade, and that the software should be considered obsolete as-of May 15th.

https://honest.cash/dgenr8/xt-wont-follow-may-19-bch-fork-4313


Mengerian:
@dgenr8 @dagurval FYI

PR #342, opened 2019-05-08
[wg] Upgrade Testnet new activation information

Mengerian:


Mengerian:
> Should we not also add `reconsiderblock` information?

Yes, good point. I have added that info in.

PR #347, opened 2019-05-13
Segwit recovery clarifications about script execution and consequences.

joshmg:
After much discussion in the Bitcoin ABC telegram dev channel, and additional internal discussion within the Software Verde team, we evaluated the consequence of ignoring the result of a Segwit-Program's P2SH script (compared to running it as a P2SH script and checking its result normally).

I think everyone is in agreement that there is very little consequence/benefit to reverting the change. 
 Therefore, we agree it's more beneficial to ensure the specification is explicit in its intended behavior (when had already been done at least once around mid-April).  We've attempted to clarify the specification so that other implementors clearly know that the regular P2SH result of a Segwit-Program must be ignored in order to remain in sync with the protocol.  We have also briefly included justification for why this is a safe development decision.

Our proposed changes do not alter the implementation, and no code changes should be required for those following the specification's intent.  Our changes merely act as further clarification and justification for future readers.

### DIFF
--- a/spec/2019-05-15-segwit-recovery.md
+++ b/spec/2019-05-15-segwit-recovery.md
@@ -30,7 +30,7 @@ Equivalently, a witness program can be identified by examining the length and th
 * The first byte is 0x00, or in the range 0x51 – 0x60. (OP_0, or OP_1 – OP_16).
 * The second byte is equal to to the redeem script byte-length, minus two.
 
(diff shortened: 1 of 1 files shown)


markblundeberg:
Looks good to me.

Mengerian:
Looks fine to me.

@floriansengel Could you also have a look and see if you like this addition to the explanation? Thanks.

joshmg:
We updated the changes to use single spaces to remain consistent with the rest of the document.

floriansengel:
Yup, seems great. Nice job @joshmg and team!

PR #391, opened 2019-08-16
Add new Upgrade Testnet information

Mengerian:
Information for the upgrade testnet for the Nov 15, 2019 upgrade.


Mengerian:
Work in progress, will add more info as it becomes available before merging this in.

PR #392, opened 2019-08-17
Add end-of-fundraiser thank you message

Mengerian:
My mediocre web design skills are evident here, so if someone can fix up the code, and make it look nicer overall, that would be great. The image could also be smaller, but I don't know how to change that.


merc1er:
#393

PR #398, opened 2019-08-22
[spec] Add more explanation of Automatic Replay Protection

Mengerian:
Many people get confused by this portion of the specification. This adds some explanation of the purpose of the automatic replay protection that can hopefully help people understand it more easily.

### DIFF
--- a/spec/2019-11-15-upgrade.md
+++ b/spec/2019-11-15-upgrade.md
@@ -40,6 +40,11 @@ Details can be found in the [full specification: 2019-11-15-minimaldata.md](2019
 
 ## Automatic Replay Protection
 
+The purpose of Automatic Replay Protection is to serve as a full node version-deprecation mechanism. It is intended to cause 
+full validating nodes which do not upgrade, to automatically separate themselves from the main network after the next
+upgrade on 15 May 2020. Nodes which implement the next upgrade will remove this automatic replay protection, and thus all regular
+wallets can continue using the default ForkID with no change to follow the main upgraded chain.
+
 When the median time past [1] of the most recent 11 blocks (MTP-11) is less than UNIX timestamp 1589544000 (May 2020 upgrade)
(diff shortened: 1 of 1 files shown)


Mengerian:
This also changes the actual specification to remove the explicit value, since no implementations actually used the value specified.

Mengerian:
@bitjson Please also have a look, I used some of your words from the Telegram conversation where we discussed this.

bitjson:
Much better, thanks!

spec/2019-11-15-upgrade.md
```diff
@@ -40,6 +40,11 @@ Details can be found in the [full specification: 2019-11-15-minimaldata.md](2019
 
 ## Automatic Replay Protection
 
+The purpose of Automatic Replay Protection is to serve as a full node version-deprecation mechanism. It is intended to cause 
+full validating nodes which do not upgrade, to automatically separate themselves from the main network after the next
```
bitjson: fully-validating*

bitjson:
That's a very helpful description 👍

spec/2019-11-15-upgrade.md
Mengerian: OK, I will make that change. Thanks for the review :)

BigBlockIfTrue:
> This also changes the actual specification to remove the explicit value, since no implementations actually used the value specified.

My confusion has now increased. If this is a consensus rule, and every implementation picks a different arbitrary value, won't that cause chain splits?

BigBlockIfTrue:
@imaginaryusername My confusion is caused because the text says:
> Bitcoin Cash full nodes implementing the Nov 2019 consensus rules SHOULD enforce the following change
instead of:
> Full nodes intending to fork away from Bitcoin Cash in Nov 2019 SHOULD enforce the following change

BigBlockIfTrue:
Brain

PR #409, opened 2019-10-16
Bump Bitcoin ABC version number

Mengerian:
Point to latest version in Upgrade notice


Mengerian:
Made the ABC version non-specific, as suggested by @jasonbcox

PR #411, opened 2019-10-18
Fixed docker-compose config

jasonbcox:
This patch does the following:
* Makes version number explicit so that certain versions of docker-compose do not puke on this config.
* Makes the service name explicit so this docker-compose can be used in a multi-project deployment environment without unnecessary changes.
* Cleans up unnecessary fields.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,12 +1,9 @@
-version: '2'
-
+version: '2.4'
 services:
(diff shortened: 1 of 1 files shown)


zquestz (approved):

jasonbcox:
Closing this to clear my open PR queue.  It's quite stale anyway.

PR #422, opened 2019-11-15
Set banner to upgrade complete

Mengerian:


_includes/home/header.html
```diff
@@ -2,66 +2,16 @@
...
       <h4 class="panel-title">
           <a>
             <ion-icon name="checkmark"></ion-icon>
             {% t 'networkupgrade.completedescription' %}
           </a>
       </h4>
-    </div>
   </div>
-</div> -->
+  </div>
```
jasonbcox: is this line duplicated?

Mengerian: Yeah, I guess there was a superfluous <div> left in there...

Mengerian: Yeah, I guess there was a superfluous <div> left in there...

PR #435, opened 2019-12-18
Change video link

Mengerian:
Add link to new video


dagurval:
That new link is broken. Is it the same video?

jasonbcox:
The video will be available in a few hours. Hang tight.

PR #440, opened 2020-01-20
Fix specs

merc1er:
What was updated:

- Restored bootstrap list styles
- Added `out of date` flags to some specs
- Display 2019 specs
- Removed specs submenu

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -181,8 +181,7 @@ learn:
     title: "tutorials"
     description: "Step by step guides to take you from basics to professional grade apps"
 specs:
-  title: "Bitcoin Cash Developer Portal"
-  description: "The following is a topical list of Bitcoin Cash Development Workgroups. Each group should be focused around a specific subject. They may exist for a limited time, or be of ongoing interest."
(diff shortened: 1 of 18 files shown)


Mengerian (approved):
Thanks @merc1er This is a big improvement from before, I noticed a few problems still though.

I did notice some things that can still be improved further:
 - [ ] The bulleted lists aren't indented (in fact they seem to stick left into the margin), is it possible to make them indented, and sub-bullets indented even more?
 - [ ] I'm not sure what you mean about the specs submenu, but the menu at the top is still screwed up (eg, clicking on "Start Here" leads to a non-existent page).
 - [ ] The relative links don't work, for example, look at /spec/2019-11-15-upgrade.html and the links to the other parts of the spec. It turns the links into links to .md files that don't exists (eg. /spec/2019-11-15-schnorrmultisig.md)
 - [ ] For the menu with the May 2019 specs, could you just change the order so that the "Network Upgrade Specification" is first? this is the overall spec, so it should always be first first.

merc1er:
I will merge this for now and submit a new PR with the changes above.

PR #447, opened 2020-02-20
Add reminder banner for May 2020 Upgrade

Mengerian:


Mengerian:
> Code looks good to me.
> Do we want to wait a little to have more compatible nodes listed? Not sure how soon (or late) other nodes will be ready.

It doesn't seem like there will be any compatible nodes anytime soon, so I'd be inclined to merge it now, and add more info as it becomes available.

I will also need to add in the Testnet and Specification info. I'm working on those, so they should hopefully be ready sooner than later.

_includes/home/header.html
```diff
@@ -1,3 +1,45 @@
...
+          </a>
+<!-- 
+          <h4>{% t 'networkupgrade.additional' %}:</h4>
+          <a href="https://github.com/bitcoincashorg/bitcoincash.org/blob/master/spec/2019-11-15-upgrade.md" title="Upgrade Specification" target="_blank">
+              <ion-icon name="link"></ion-icon> {% t 'networkupgrade.specification' %}
+          </a>
+          <a href="https://github.com/bitcoincashorg/bitcoincash.org/blob/master/workgroups/wg-testing/2019-11-15_upgrade_testnet.md" title="Testnet Information" target="_blank">
+           <ion-icon name="link"></ion-icon> {% t 'networkupgrade.testnet' %}
+          </a>
+-->
```
jasonbcox: It's best to remove this blob of commented code and open a PR for it when it's ready.

_includes/home/header.html
```diff
@@ -1,3 +1,45 @@
...
+        </div>
+      </div>
+    </div>
+  </div>
+  </div>
```
jasonbcox: Fix indentation

BigBlockIfTrue:
There needs to be some specific warning that Bitcoin ABC 0.21 in default configuration enforces an additional consensus rule that is not part of the Bitcoin Cash protocol and hence may fork itself off the Bitcoin Cash chain.

merc1er:
Agreed with @BigBlockIfTrue. 

@Mengerian I meant wait for an alternative node so that people can make up their own mind before upgrading their software.

BitcoinCash.org should remain neutral shouldn’t it?

ftrader:
> hence may fork itself off the Bitcoin Cash chain.

I agree with @BigBlockIfTrue that this risk is real, and there should be a warning about it.

merc1er:
@Mengerian @jasonbcox

PR #444, opened 2020-02-15
[spec] Add specification for May 2020 Upgrade

Mengerian:

PR #450, opened 2020-02-28
Add BCHNode to list of upgrade-compatible nodes

BigBlockIfTrue:


sickpig:
@Mengerian what about showing the same urge it has been used in merging #444 to sneak in IFP in the May 2020 specification well after Feb 15th feature freeze?

Mind you that Feb 15th deadline is for code, spec should be finalize in time for devs to implement the features in their code base, eg well before feature freeze deadline.

Mengerian:
Given that BCHN was created with the express purpose of being incompatible with the IFP, which was included in the Bitcoin ABC upgrade release, it seems a bit strange to add it to the list of "compatible nodes".

But regardless of that, there are a couple reasons to reject this.

Firstly, it's a very new project. Typically we wait for a few months until adding new node projects. A good example is BCHD, which I actually asked @cpacia about adding early on, and he suggested I wait a while since it was such a new project at the time. So just for that reason, it would be better to wait a while to see how the project pans out before adding it to the site.

Secondly, this is a risk for people to use for mining. Again BCHD is a good counter-example. The BCHD project clearly discourages use of their node for mining, since there is a risk it could produce invalid blocks. BCHN, on the other hand, has been promoting it's node as a "drop in replacement" for ABC, with no mention (that I'm aware of) of the risks of mining with it. If IFP were to activate, then mining with BCHN is risky for the miner, as it will produce blocks orphaned by Bitcoin ABC which comprises the majority of the network. If there were clear messaging that the node is not to be used for mining, then it would be more justifiable. (Although the newness of the project would still be an issue).

sickpig:
> Given that BCHN was created with the express purpose of being incompatible with the IFP, which was included in the Bitcoin ABC upgrade release, it seems a bit strange to add it to the list of "compatible nodes".

you meant compatible with "Bitcoin Cash" right? I'm sure that the use of "Bitcoin ABC" has to be some case of Freudian slip.

> Firstly, it's a very new project. 
> Secondly, this is a risk for people to use for mining.

BCHN = ABC 0.21.0 - IFP 

Are you implying that ABC 0.21 is risky for mining?

Frankly I found your motivations just excuses for not listing BCHN on bitcoincash.org, of course it's in your power to do so since you have been "nominated" as the gatekeeper for its content by the owner of the domain. 

But I rather like you stating your decisions w/o pretending to be the result of a rational / democratic process.

Mengerian:
> BCHN = ABC 0.21.0 - IFP
> 
> Are you implying that ABC 0.21 is risky for mining?

Yeah, mining on ABC with -enableminerfund=0 would be similarly risky.
 
> But I rather like you stating your decisions w/o pretending to be the result of a rational / democratic process.

Well I don't think these decisions should be democratic, but I hope they are rational.

sickpig:
> > BCHN = ABC 0.21.0 - IFP
> > Are you implying that ABC 0.21 is risky for mining?
> 
> Yeah, mining on ABC with -enableminerfund=0 would be similarly risky.

Independently of your reasoning behind your decision to deem it risky, I guess that the rational thing to do is for all miners, exchanges and users in general to stop using ABC since it is the entity that introduced the risk into the system.

> > But I rather like you stating your decisions w/o pretending to be the result of a rational / democratic process.
> 
> Well I don't think these decisions should be democratic, but I hope they are rational.

FWIW neither of those adjectives applies to the process.

The fact alone that you are saying that the process shouldn't be democratic is pretty telling and I wonder what is the reason of going through public review for PRs that modify specifications. Why not just publish the spec after a round of private reviews and call it a day? 

While at it what's the reason to justify the publishing the specification in the first place?

After all you think, [as you stated in commenting](https://github.com/bitcoincashorg/bitcoincash.org/pull/453#pullrequestreview-369878336) #453, that spec is "to inform people about what is actually implemented in the code" and ABC "is used by almost all miners". Does people really need to know how ABC behave consensus wise since everyone run ABC?

PR #455, opened 2020-03-05
Add per-transaction limit to SigChecks spec

Mengerian:
Since a per-transaction limit was implemented, the spec should reflect that.

### DIFF
--- a/spec/2020-05-15-sigchecks.md
+++ b/spec/2020-05-15-sigchecks.md
@@ -1,10 +1,10 @@
 ---
 layout: specification
 title: 2020-MAY-15 script SigChecks counting and limiting specification
-date: 2020-02-10
+date: 2020-03-05
 category: spec
 activation: 1589544000
-version: 0.1 (DRAFT)
+version: 0.2 (DRAFT)
 author: Mark B. Lundeberg
 ---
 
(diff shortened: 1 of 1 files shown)


spec/2020-05-15-sigchecks.md
```diff
@@ -47,6 +47,12 @@ Here, max_Blocksize refers to the consensus limit that is enforced on the full s
 
 Currently, `max_BlockSize` = 32000000 so the maximum allowed `block_SigChecks` would be 226950 for all blocks. However, in future block size increases, it should be assumed that the SigChecks limit increases proportionally.
 
+## Per-transaction limits (consensus rule)
+
+After activation, any transaction where the total number of SigChecks accumulated (during all script executions - scriptSig, scriptPubKey, and P2SH redeemScript - in all inputs excepting coinbase) violates the following limit shall be rejected:
+
+    transaction_SigChecks <= 3000
```
sickpig: Any rationale about this limit?

spec/2020-05-15-sigchecks.md
```diff
@@ -152,18 +158,10 @@ The choice of 141 bytes/SigCheck for a block is ~4x times more aggressive than t
 A histogram of historical block densities is plotted below:
 ![Block sigchecks density historically (up to mid-2019)](2020-05-15-sigchecks-plotblocks.png)
 
-
-## Why no per-transaction limits?
-
-The per-block consensus limit is enough to track all validation happening in blocks, and the per-input standard limit is satisfactory for limiting mempool validation time.
```
BigBlockIfTrue: What new insight led to a change of mind?

spec/2020-05-15-sigchecks.md
jasonbcox: Providing another lever that can be tweaked (or not) when bumping the block size is just good engineering.  Without a per-tx limit, any adjustment to the block size limit would need to implicitly take into account the effect of the number of sigchecks that a malicious transaction could consume before becoming problematic (ie consuming an inordinate amount of resources to verify). Conversely, providing such a limit provides a less-encumbered path to bumping the block size, when such a decision needs to be made.

PR #456, opened 2020-03-05
Remove Bitprim node

Mengerian:
This should have been done a while ago. It is no longer an active project.


Mengerian:
Thanks @merc1er

PR #458, opened 2020-03-07
Bump Bitcoin ABC version in website banner

Mengerian:


merc1er:
@Mengerian looks good, but just double checking: are the releases now done through GitHub instead of bitcoinabc.org website?

Mengerian:
> @Mengerian looks good, but just double checking: are the releases now done through GitHub instead of bitcoinabc.org website?

The release is in both places, in fact there's a link to the release on bitcoinabc.org from the Github release.

I just put the Github release for this one because I thought it would be nice for people to be able to see the release notes.

PR #475, opened 2020-04-20
Added Crypto.com to exchanges

brianMCO:
Hi,
We would like to list Crypto.com Exchange in the exchange list at bitcoincash.org. Thanks!


brianMCO:
Thanks!

PR #482, opened 2020-05-06
Update ABC version in upgrade page

Mengerian:
This was missed in https://github.com/bitcoincashorg/bitcoincash.org/pull/481


Mengerian:
Rebased on master

PR #484, opened 2020-05-06
Add BCH Unlimited to upgrade compatible implementations

sickpig:

PR #485, opened 2020-05-07
Add Bitcoin Cash Node to upgrade compatible implementations

sickpig:
Built on top of #484 to avoid merge conflicts

### DIFF
--- a/_includes/home/header.html
+++ b/_includes/home/header.html
@@ -29,6 +29,9 @@
            <a href="https://bitcoinunlimited.info/download/" title="BCH Unlimited 1.8.0.0" target="_blank">
                <ion-icon name="link"></ion-icon> BCH Unlimited 1.8.0.0
            </a>
+           <a href="https://bitcoincashnode.org/en/download.html" title="Bitcoin Cash Node 0.21.x" target="_blank">
+               <ion-icon name="link"></ion-icon> Bitcoin Cash Node 0.21.x
+           </a>
            <h4>{% t 'networkupgrade.additional' %}:</h4>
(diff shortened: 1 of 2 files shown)


sickpig:
Hey @Mengerian @jasonbcox,

I just discovered that the owner of the bitcoincash.org 
(@deadalnix) thinks that BCHN is ["very hostile and threaten to fork"](https://youtu.be/lHGnfZWu1eU?t=10919) and for that reason it won't be listed among the list of compatible clients.

Just wanna to point out that this is merely a political play in my book. Both BCHD and BCH Unlimited are exactly the same of BCHN when it comes to the set of features needed for today's upgrade.

All the 3 clients implement all the needed stuff but IFP. So there's no technical reason to not include BCHN.

It had been better to say it loud and clear since the beginning that there wasn't any intention to merge this PR rather than keep it stalled here w/o 0 feedback. 

Not very professional if you ask me.

georgedonnelly:
> Not very professional if you ask me.

Andrea, we are doing the best job we can with the limited resources available to us. I would counter that it is also not very professional to throw rocks at guys who are working past their limits, especially on the day of a major network upgrade.

Thank you everyone for your patience. We aim to respond to all requests within a reasonable time frame.

jasonbcox:
I agree, this is poor form from @sickpig. This behavior from you is going to get you ignored by everyone who is focusing on hard work rather than wasting their time with political games about who gets listed on a website.

BCHN needs to grow up before a PR like this will be considered.

ftrader:
> The project does not have any degree of prolonged activity and provides questionable, if any, value to Bitcoin Cash.

Keep merging our improvements while you maintain we do not provide any value, Jason.

pokkst:
I sense a lot of projection coming from ABC. Hmmmm.

PR #486, opened 2020-05-07
Adds Knuth 0.2.0 as a compatible node for 2020-May

fpelliccioni:


sickpig:
@jasonbcox Knuth is Bitprim node re-branded.

fpelliccioni:
> How long has this node project been active? Is there a case for Knuth running in the wild as a stable implementation?

Hey Jason,
It is active since the beginning of 2020 by the name of Knuth, if we count its predecessor, Bitprim, it is active in Bitcoin Cash from the beginning, we started supporting BCH at the same time as ABC.

jasonbcox:
Is this a fork of the same code base or just rebranded in spirit?

fpelliccioni:
> Is this a fork of the same code base or just rebranded in spirit?

I do not understand your question, but...
Bitprim was discontinued last year, so it no longer exists.
Knuth was born as a fork of Bitprim but contains several performance improvements and support for the current version of the protocol.

PR #494, opened 2020-06-23
Remove Yenom wallet

Mengerian:
The project appears to be defunct

### DIFF
--- a/_includes/wallets/wallets.html
+++ b/_includes/wallets/wallets.html
@@ -78,7 +78,7 @@
 <section id="wallets-section" class="space--none text-center">
   <div class="container wallets">
     <div class="row featured-wallets">
-      <div class="col-sm-4 col-md-4 vertical-align to-animate">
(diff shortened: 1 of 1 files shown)


georgedonnelly:
I emailed the google play store contact address 21 Jun, almost 48 hours ago, no reply yet.

merc1er:
Thanks! Let’s remove it ASAP to avoid having more people lose funds!

Mengerian:
For the record, here's an announcement by Yenom that the wallet is shut down: https://twitter.com/yenom_wallet/status/1285016199516925952

English via Google Translate:
"Thank you for using Yenom.
The service will be terminated as of the end of December 2020.

Since both iOS and Android services will be terminated, please move to another company's wallet or exchange."

Mengerian:
Actually, they did have an English Tweet also: https://twitter.com/Yenom_wallet_en/status/1285016979103862785

PR #499, opened 2020-08-01
Update docker build do include gatsby proj, serve both from nginx

DanielLyons178:
Include gatsby build in master docker image, serve v2 by nginx if exists else serve v2

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,10 +17,8 @@ FROM node:14.5.0-buster-slim as build2
 
 WORKDIR /app
 RUN npm -g install gatsby-cli && gatsby telemetry --disable
-ADD ./v2/package.json ./package.json
-ADD ./v2/package-lock.json ./pakcage-lock.json
-RUN npm install --production --frozen-lockfile --non-interactive
 ADD ./v2/. ./
+RUN npm install --production --frozen-lockfile --non-interactive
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -12,7 +12,19 @@ COPY . .
 RUN mv _config-${APP_ENV}.yml _config.yml
 RUN bundle exec jekyll build
 
+
+FROM node:14.5.0-buster-slim as build2
+
+WORKDIR /app
+RUN npm -g install gatsby-cli && gatsby telemetry --disable
+ADD ./v2/package.json ./package.json
+ADD ./v2/package-lock.json ./pakcage-lock.json
```
jasonbcox, line 21: Typo in destination file

Dockerfile
```diff
@@ -12,7 +12,19 @@ COPY . .
...
+RUN npm install --production --frozen-lockfile --non-interactive
+ADD ./v2/. ./
```
jasonbcox, line 23: Since you're adding all files from v2, why not add them before npm install? Then you don't need to be explicit regarding the package files.

jasonbcox: I realized I'm actually wrong on this one. This busts the cache on every change and makes the build super slow.

PR #503, opened 2020-08-04
Feature/language dropdown

JakeAngell:


JakeAngell:
Now that there are two dropdown components this can be abstracted, I will do this as a separate refactoring commit in the coming days

JakeAngell:
Will need to tweak the design/implementation to make the mobile menu more accessible.

On devices with low height the language dropdown can overflow

v2/src/components/header/navBarItems.js
```diff
@@ -9,12 +11,31 @@ const communityDropdownLinks = [
...
+
+const languageDropdown =(<><World></World><div style={{paddingLeft:'5px'}}>Language</div></>);
 
 const navBarItems =[{index:1, text:"Get started", href:"/start-here.html"},
      {index:2, text:"Wallets", href:"/wallets.html"},
      {index:3, text:"Logos", href:"/graphics.html"},
      {dropdown: <CommunityDropdown links={communityDropdownLinks} index={4} key={4} text={"Community"}></CommunityDropdown>, mobileDropdown: <MobileCommunityDropdown links={communityDropdownLinks} text={"Community"} key={4}></MobileCommunityDropdown>},
-     {index:5, text:"About", href:"/faq.html"}];
+     {index:5, text:"About", href:"/faq.html"},
+     {dropdown: <LanguageDropdown links={languageDropdownLinks} index={6} key={6} text={languageDropdown}></LanguageDropdown>, mobileDropdown: <MobileCommunityDropdown links={languageDropdownLinks} text={languageDropdown} key={6}></MobileCommunityDropdown>}];
```
jasonbcox, line 38: Nit: You can avoid the removal line in this patch by putting `},` at the end of the line and putting `];` on its own line. This makes it easier to review.

Ideally we'd have a linter do this.

JakeAngell: Added cursor and hover styling. 

Will add a linter in the upcoming refactor PR to avoid issues like this in the future.

PR #521, opened 2020-08-17
Revert "Fix indentation in layout.js"

nhutran1201:
Reverts bitcoincashorg/bitcoincash.org#515


jasonbcox:
Why is this reverted? I don't see the point.

PR #530, opened 2020-08-17
Add translation support via transifex

deadalnix:
Enable translation of the legalese in French. Checking the locale in layout.js indeed changes the text.


deadalnix:
I will have to rework the file format for these translations, so that it matches what transifex expects.

deadalnix:
OK, transifex is up and running now

PR #508, opened 2020-08-13
[spec] Add specification for November 2020 Upgrade

uahf:
Included:

- Main upgrade document
- Full ASERT (aserti3-2d) specification

The upgrade proposal has been accepted by all Bitcoin Cash clients and enjoys widespread ecosystem support for November.

The original versions of these files come from and are maintained here:

https://gitlab.com/bitcoin-cash-node/bchn-sw/bitcoincash-upgrade-specifications/-/tree/master/spec

They are also visible via https://upgradespecs.bitcoincashnode.org/ .

Issues with the specification can be raised via

https://gitlab.com/bitcoin-cash-node/bchn-sw/bitcoincash-upgrade-specifications/-/issues/new

It is also possible to submit change requests against the specification directly by creating a Merge Request on the repository at https://gitlab.com/bitcoin-cash-node/bchn-sw/bitcoincash-upgrade-specifications/ .

Discussion on the upgrade spec is welcome via open-invite Slack at 

https://join.slack.com/t/bitcoincashnode/shared_invite/zt-egg3c36d-2cglIrKcbnGpIQFaKFzCWA

PR #533, opened 2020-08-18
Fixed inconsistency in Python code, simplified names further

EyeOfPython:
There was an inconsistency in the Python code between `current_target` and `next_target`, which has been fixed.

Now the spec only uses three names for different blocks:

* `anchor` for the parent of the fork block, used as reference block.
* `current` for the current best block, parent of `next` block.
* `next` for the block that‘s about to be mined.

This should make it much easier to implement this algorithm for newcomers.

### DIFF
--- a/spec/2020-11-15-asert.md
+++ b/spec/2020-11-15-asert.md
@@ -65,12 +65,12 @@ The current block's target bits are calculated by the following algorithm.
 The aserti3-2d algorithm can be described by the following formula:
 
 ```
-next_target = old_target * 2**((time_delta - ideal_block_time * (height_delta + 1)) / halflife)
+next_target = anchor_target * 2**((time_delta - ideal_block_time * (height_delta + 1)) / halflife)
 ```
 
 where:
 
-- `old_target` is the unsigned 256 bit integer equivalent of the `nBits` value in
+- `anchor_target` is the unsigned 256 bit integer equivalent of the `nBits` value in
   the header of the anchor block.
 - `time_delta` is the difference, in signed integer seconds, between the
   timestamp in the header of the current block and the timestamp in the
@@ -95,7 +95,7 @@ block header.
 Python-code, uses Python 3 syntax:
 
 ```python
-def aserti3_2d(
+def next_target_aserti3_2d(
     anchor_height: int,       # height of the anchor block.
     anchor_parent_time: int,  # timestamp (nTime) of the parent of the anchor block.
     anchor_bits: int,         # 'nBits' value of the anchor block.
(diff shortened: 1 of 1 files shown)


EyeOfPython:
@markblundeberg @deadalnix @jtoomim If possible, could you review whether my simplification of the spec is correct?

jtoomim:
This repository is owned and controlled by Bitcoin ABC. Bitcoin ABC has made it clear that this repository reflects Bitcoin ABC's specification, and that comments from other implementations on consensus rules will be ignored if those comments are inconvenient to ABC's goals:

https://read.cash/@BigBlockIfTrue/this-conversation-was-marked-as-resolved-8ef84149

I do not find it worthwhile to expend effort on this when Bitcoin ABC's plan is to fork off anyway.

The specification for aserti3-2d is here: https://gitlab.com/bitcoin-cash-node/bchn-sw/bitcoincash-upgrade-specifications/-/blob/071d8f7b77b3a28bdad1a9e9fcbbac8d1adb2afc/spec/2020-11-15-asert.md

The version you have here is not equivalent to the aserti3-2d spec. You and Bitcoin ABC should not be rewriting the spec in the format you prefer. Doing so has the potential to introduce subtle errors into the spec. 

If you had feedback on the style or formatting of the spec, and would have liked to see it written differently, you should have voiced that feedback *before* August 15th, and you should have voiced that feedback on MR !7, where all of the other 311 comments of feedback were given.

https://gitlab.com/bitcoin-cash-node/bchn-sw/bitcoincash-upgrade-specifications/-/merge_requests/7

EyeOfPython:
Hey thanks @jtoomim for the heads up.

BigBlockIfTrue:
@EyeOfPython, you can find information on how to contribute to the ASERT spec in PR #508 here. However, at this stage we should indeed limit ourselves to minor edits, for the reason Jonathan stated.

The language of the pseudocode has already been discussed during writing so it is very late to change that now. Replacing the pseudocode with Python is also unnecessary since there is already a Python reference implementation attached to the spec.

jtoomim:
I'm going to repeat this, because I think it is important:

> The version you have here is not equivalent to the aserti3-2d spec.

_There is already at least one subtle bug in this spec._ I strongly recommend that you abandon your efforts at rewriting a frozen spec, and simply use the same spec that the rest of the BCH community is using.

I am willing to review _implementations_ to check for conformance with [the spec](https://gitlab.com/bitcoin-cash-node/bchn-sw/bitcoincash-upgrade-specifications/-/blob/071d8f7b77b3a28bdad1a9e9fcbbac8d1adb2afc/spec/2020-11-15-asert.md), but you should not be rewriting or modifying someone else's spec unless your goal is to make a non-conformant derivative.

If you wish to host the copy of the spec on your repository, that seems fine with me. However, it should be a bit-identical copy, not a derivative.

PR #540, opened 2020-08-19
Decorate links with the proper locale

deadalnix:
This required to mess up the language menu because it needs to be special cased now so that it doesn't get decorated.


deadalnix:
Refactored this toward a much nicer way to do it.

jasonbcox (dismissed):

PR #546, opened 2020-08-20
Updates to nav bar

johnkuney:
Replaced white logo with color versoin and made clickable
Style edits to price ticker
Rearrangment of nav items


v2/src/components/liveprice/live-price-widget.js
```diff
@@ -21,14 +21,11 @@ const LivePriceWidget = ({ currentPrice, ticker, url }) => {
   return (
     <a className={livePricingStyles.widgetContainer} href={url}>
       <div className={livePricingStyles.currentPrice}>{currentPrice}</div>
-      <div className={livePricingStyles.tickerContainer}>
         <div
           className={livePricingStyles.ticker}
-          style={{ color: theme.primary_dark }}
         >
           {ticker}
         </div>
```
deadalnix, line 28: make sure that you run `npm run format` so that indentation gets fixed up.

v2/src/layout/header.js
```diff
@@ -46,67 +46,116 @@ const languageDropdownLinks = Object.entries(locales).map(([_, locale]) => {
...
+    },
     {
       text: (
         <>
           <World />
-          <span style={{ paddingLeft: "5px" }}>
+          {/* <span style={{ paddingLeft: "5px" }}>
             <fbt desc="Language selector menu">Language</fbt>
-          </span>
+          </span> */}
```
deadalnix, line 145: Don't leave code commented out.

v2/src/layout/header.js
```diff
@@ -46,67 +46,116 @@ const languageDropdownLinks = Object.entries(locales).map(([_, locale]) => {
...
       links: languageDropdownLinks,
     },
   ]
 
+  const logoLink = [
+    {
+      text: fbt("Logo Link", "Top 'logo' link"),
+      href: "/",
+    }
+  ]
```
deadalnix, line 157: That seem completely redundant. Why?

v2/src/layout/header.js
```diff
@@ -46,67 +46,116 @@ const languageDropdownLinks = Object.entries(locales).map(([_, locale]) => {
...
   const communityDropdownLinks = [
     {
-      text: fbt("Services", "Communty menu 'services' link"),
+      text: fbt("r/BTC", "Communty menu 'rBTC' link"),
+      href: "https://www.reddit.com/r/btc/",
+    },
+    {
+      text: fbt("r/BitcoinCash", "Communty menu 'rBitcoinCash' link"),
+      href: "https://www.reddit.com/r/Bitcoincash/",
+    },
```
deadalnix, line 55: No need for both. Considering r/btc has turned into clown world, we probably want to go for r/bitcoincash only .

It is not clear to me that we want social media that high up. This is expensive real estate and comes at the price of getting links such as "get started" in a menu rather than right there.

Also, do we rally wnt to send things such as r/bitcoincash , Telegram, Facebook and so on for translation? I'm not convinced this is worth it, because these are brands.

v2/src/layout/header.js
```diff
@@ -46,67 +46,116 @@ const languageDropdownLinks = Object.entries(locales).map(([_, locale]) => {
...
+
+  const resourcesDropdownLinks = [
+    {
+      text: fbt("Get Started", "Resources menu 'Get started' link"),
+      href: "/start-here.html",
+    },
+    {
+      text: fbt("History", "Resources menu 'History' link"),
       href: "/services.html",
     },
```
deadalnix, line 82: You link to services.html twice.

v2/src/layout/header.js
```diff
@@ -152,11 +201,22 @@ const Header = () => {
...
+                  data-sal-easing="ease"
+                  className={headerStyles.logo}
+                  to={logoLink.href}
+                >
+                  <img
+                    src={logo}
+                    alt="bitcoincashlogo"
+                  />
+                </Link>
+            )}
```
deadalnix, line 218: Create an array with one item and then mapping over it here seems completely redundant.

deadalnix (changes requested):
I have a few comments on the code, but also a more generic advice.

This is doing two things:
1/ Changing the logo in the nav bar so that is brings us back home + style.
2/ Changing the content of the menus.

By submitting both at the same time, you will impedes on your ability to iterate quickly in various ways:
1/ Larger patches are harder to review, and the larger the patch, the slower the review cycle will be.
2/ Any change required on one of the elements will prevent the other element from being merged.
3/ The longer the patch stays up, the more likely it is that you'll run into merge conflicts. Combined with 1 and 2, this compounds the slowdown.
4/ Large patches are harder to revert - for the same reason they cause more merge conflicts. This leads to more reluctance from reviewer to accept them, because the cost of error is greater.
5/ While this does not really apply to this patch, you might find yourself in a position where the people knowledgeable enough about the various parts of your patches are not the same, in which case the slowdown will be even more dramatic.

I recommend you red [this](https://medium.com/@kurtisnusbaum/large-diffs-are-hurting-your-ability-to-ship-e0b2b41e8acf) if you want to know more.

v2/src/components/liveprice/live-price-widget.js
johnkuney: Okay np ran that

v2/src/layout/header.js
johnkuney: removed

v2/src/layout/header.js
johnkuney: Yeah not sure what I was doing there...fixed

v2/src/layout/header.js
johnkuney: Good points I thought maybe community channels were important to promote high up? But can the case for maybe leaving them in the footer. The resource tab was also getting quite long so I split out developer related items to its own tab which I think makes more sense and is a valid section to give prime spot to

v2/src/layout/header.js
johnkuney: There is no histrory page yet so was an placeholder but I've replaced with the likely url

v2/src/layout/header.js
johnkuney: Yeah thats unnecessary, fixed.
And overall I understand more and more now your work flow philosophy and its benefits. Sorry for taking a few times to catch up, but I guess I've been guilty of doing large changes in my past work. I think you'd be shocked at the size of diffs that happen at bitcoin.com sometimes lol. I will do my best to compartmentalize changes going forward.

v2/src/layout/header.js
```diff
@@ -152,11 +171,17 @@ const Header = () => {
...
+                data-sal="slide-down"
+                data-sal-delay={200}
+                data-sal-duration="1000"
+                data-sal-easing="ease"
+                className={headerStyles.logo}
+                to="/"
+              >
+                <img src={logo} alt="bitcoincashlogo" />
+              </Link>        
+            <div className={headerStyles.divider}></div>
```
deadalnix, line 184: Can you extract that part of the PR in its own PR?

deadalnix (changes requested):

deadalnix (changes requested):
Please separate the different part of the PR.

v2/src/layout/header.module.scss
```diff
@@ -132,4 +132,4 @@ header {
   .logo {
     margin-right: 14px;
   }
-}
+}
```
deadalnix, line 135: Revert (or just run npm run format and it'll do it for you).

deadalnix (changes requested):
If you extract the ticker restyling, then this can also be merged directly, and that would leave this patch with only the decision about how the menu are structured. This would leave more room for someone like christabel to review that part and allow to move on other parts of the code.

v2/src/layout/header.js
```diff
@@ -44,63 +44,78 @@ const languageDropdownLinks = Object.entries(locales).map(([_, locale]) => {
...
     },
     {
-      text: fbt("Community", "Commnity menu"),
-      links: communityDropdownLinks,
+      text: fbt("Roadmap", "Top 'roadmap' link"),
+      href: "/roadmap.html",
     },
-    { text: fbt("About", "Top 'about' link"), href: "/faq.html" },
     {
       text: (
```
deadalnix, line 116: You can do text: <World />

No need for a <></> wreapper is there is only one jsx element.

johnkuney:
Okay good point. I've backed out those remaining changes so now its just the nav items. Can revisit when Christabel gets back on that

v2/src/layout/header.js
johnkuney: Good call, fixed

johnkuney:
Changing the nav to match the current site layout so there is not a difference between them in the interim when both sites are live. Its a weird UX if the nav changes from page to page. I've stashed the changes I had made and can add back in when it makes sense

v2/src/layout/header.js
```diff
@@ -46,48 +46,47 @@ const languageDropdownLinks = Object.entries(locales).map(([_, locale]) => {
...
     {
-      text: fbt("Developer Portal", "Communty menu 'developper portal' link"),
+      text: fbt("Developer Portal", "Community menu 'developper portal' link"),
       href: "/developers.html",
     },
-    {
-      text: fbt("Whitepaper", "Communty menu 'whitepaper' link"),
-      href: "/bitcoin.pdf",
-      localize: false,
-    },
```
deadalnix, line 76: You can keep the whitepaper

v2/src/layout/header.js
```diff
@@ -46,48 +46,47 @@ const languageDropdownLinks = Object.entries(locales).map(([_, locale]) => {
...
+      text: fbt("Start Here", "Top 'Start here' link"),
       href: "/start-here.html",
     },
     {
       text: fbt("Wallets", "Top 'wallets' link"),
       href: "/wallets.html",
     },
     {
-      text: fbt("Explorer", "Link to the block explorer"),
-      href: "https://explorer.bitcoincash.org/",
```
deadalnix, line 90: IMO it is more valuable to have the explorer there than logos. Can we modify the old layout instead?

deadalnix (changes requested):

deadalnix (approved):

deadalnix:
Fantastic. I'm not so familiar with the legacy site, so I'll let @jasonbcox have a look before merging.

PR #548, opened 2020-08-20
Make nav logo link to home

johnkuney:


v2/src/layout/header.js
```diff
@@ -152,11 +152,17 @@ const Header = () => {
...
+                data-sal="slide-down"
+                data-sal-delay={200}
+                data-sal-duration="1000"
+                data-sal-easing="ease"
+                className={headerStyles.logo}
+                to="/"
+              >
+                <img src={logo} alt="bitcoincashlogo" />
+              </Link>        
+            <div className={headerStyles.divider}></div>
```
jasonbcox, line 165: For empty tags, just terminate the first tag at the end like this:
`<div className={headerStyles.divider} />`

v2/src/layout/header.js
```diff
@@ -152,11 +152,17 @@ const Header = () => {
...
-            <img
-              src={logo}
-              className={headerStyles.logo}
-              alt="bitcoincashlogo"
-            />
+              <Link
+                data-sal="slide-down"
+                data-sal-delay={200}
+                data-sal-duration="1000"
+                data-sal-easing="ease"
```
jasonbcox, line 159: This confuses me a little. Is this an animation on the link? should this be one container level up from this?

v2/src/layout/header.js
johnkuney: ah for sure. fixed

v2/src/layout/header.js
johnkuney: Yeah could point I've moved it up one level so the price widget animates with it

PR #550, opened 2020-08-20
Adding fbt to large feature section

johnkuney:
Added fbt to this section and removed the global strings file as its no longer needed


v2/src/components/home/home-large-feature.js
```diff
@@ -18,18 +9,44 @@ import Anywhere from "../../assets/icons/anywhere.svg"
...
+      )}
       featuredPoints={[
-        { icon: <PhoneIcon></PhoneIcon>, text: EN_LARGE_FEATURE_POINTONE },
-        { icon: <TinyFees></TinyFees>, text: EN_LARGE_FEATURE_POINTTWO },
-        { icon: <SentFast></SentFast>, text: EN_LARGE_FEATURE_POINTTHREE },
+        {
+          icon: <PhoneIcon />,
+          text: fbt(
+            "Payments go directly from one person to another.",
+            "Large feature's point one"
```
deadalnix, line 21: These description are not going to mean much to a translator. They need to figure out where the string fits in the website to be able to translate understanding the context.

deadalnix (changes requested):

v2/src/components/home/home-large-feature.js
johnkuney: Ah yes those could be improved. Done

v2/src/components/home/home-large-feature.js
```diff
@@ -1,15 +1,6 @@
...
-  EN_LARGE_FEATURE_POINTONE,
-  EN_LARGE_FEATURE_POINTTWO,
-  EN_LARGE_FEATURE_POINTTHREE,
-  EN_LARGE_FEATURE_CTA_BODY,
-  EN_LARGE_FEATURE_CTA_BTN,
-  EN_LARGE_FEATURE_SUBTITLE,
-  EN_LARGE_FEATURE_TITLE,
-} from "../../global/strings"
+import fbt from "fbt"
 import PhoneIcon from "../../assets/icons/directpayment.svg"
```
deadalnix, line 4: While you are at it, removing the ../../ before assets is a nice cleanup. In this patch or another. We properly resolve from src.

deadalnix (approved):

v2/src/components/home/home-large-feature.js
johnkuney: For sure

PR #560, opened 2020-08-26
Add a way to specify which v2 pages get built and deployed

jasonbcox:
This provides an easy way to turn v2 pages on and off depending on environment. Default behavior is:
* `npm run build` or `npm run develop` will build all v2 pages for easy local development
* `make run` is not affected (v1 only)
* `make run APP_VERSION=2` is not affected (dev environment which includes all v2 pages)

Setting environment is easy: `make run APP_VERSION=2 APP_ENV=prod`
Similarly: `APP_ENV=prod npm run build`
APP_ENV=beta may also be used.

As pages become completed, different environments can have pages toggled and CI will deploy those v2 pages accordingly.

### DIFF
--- a/v2/gatsby-config.js
+++ b/v2/gatsby-config.js
@@ -25,8 +25,8 @@ module.exports = {
     {
       resolve: "gatsby-source-filesystem",
       options: {
-        name: "images",
-        path: "src/assets/images",
+        name: "jekyll-migration",
+        path: "src/jekyll-migration",
       },
     },
     {
--- a/v2/gatsby-node.js
+++ b/v2/gatsby-node.js
@@ -1,4 +1,5 @@
 const locales = require("./src/i18n/locales")
+const path = require("path")
 
 // Execute some function for each locale of a given page
 const forEachLocale = async (page, fn) => {
@@ -17,6 +18,84 @@ const forEachLocale = async (page, fn) => {
       }
 
       await fn(page, context, localizationParams, localizedPath)
+    })
+  )
+}
+
+const createPageForEachLocale = async (createPage, page) => {
+  await forEachLocale(
+    page,
+    async (page, context, localizationParams, localizedPath) => {
+      createPage({
+        ...page,
+        path: localizedPath,
+        context,
+      })
+    }
+  )
+}
+
+const createRedirectForEachLocale = async (createRedirect, page) => {
+  await forEachLocale(
+    page,
+    async (page, context, localizationParams, localizedPath) => {
+      createRedirect({
+        fromPath: localizedPath,
+        // By default, assume paths of the form `/path/` transform to
+        // v1 paths of the form `/path.html`
+        toPath: localizedPath.replace(/\/$/, "") + ".html",
+        redirectInBrowser: true,
+        isPermanent: false,
+      })
+    }
+  )
+}
+
+exports.createPages = async ({ graphql, actions, reporter }) => {
+  const { createPage, createRedirect } = actions
+
+  // Fetch path info for all pages
+  const result = await graphql(`
+    {
+      allFile(
+        filter: {
+          sourceInstanceName: { eq: "jekyll-migration" }
+          extension: { eq: "js" }
+        }
+      ) {
+        nodes {
+          name
+          relativeDirectory
+          absolutePath
+        }
+      }
+    }
+  `)
+  if (result.errors) {
+    reporter.panicOnBuild("Error while running GraphQL query.")
+    return
+  }
+
+  await Promise.all(
+    result.data.allFile.nodes.map(async node => {
+      let originalPath = "/" + (node.name == "index" ? "" : node.name + "/")
+      if (node.relativeDirectory) {
+        originalPath = "/" + node.relativeDirectory + originalPath
+      }
+
+      const page = {
+        path: originalPath,
+        component: path.resolve(node.absolutePath),
+        context: {},
+      }
+
+      // In production, we redirect in-development v2 pages to v1 equivalent routes.
+      // In non-prod environments, we create all pages and do not redirect.
+      if (process.env.GATSBY_APP_ENV == "prod") {
+        await createRedirectForEachLocale(createRedirect, page)
+      } else {
+        await createPageForEachLocale(createPage, page)
+      }
     })
   )
 }
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 # Optionally build v2 layer
 FROM node:14.8.0-buster-slim as buildv2
 ARG APP_VERSION=1
+ARG APP_ENV=prod
 WORKDIR /work
 
 # Ensure the public dir exists for v1 builds
@@ -10,7 +11,7 @@ RUN if [ "${APP_VERSION}" -eq "2" ]; then npm -g install gatsby-cli ; fi
 COPY v2/package.json v2/package-lock.json ./
 RUN if [ "${APP_VERSION}" -eq "2" ]; then npm install --frozen-lockfile --non-interactive ; fi
 COPY v2 .
-RUN if [ "${APP_VERSION}" -eq "2" ]; then npm run build ; fi
+RUN if [ "${APP_VERSION}" -eq "2" ]; then GATSBY_APP_ENV="${APP_ENV}" npm run build ; fi
 
 # Build v1 layer
 FROM ruby:2.5.1-stretch as buildv1
--- a/v2/src/jekyll-migration/README.md
+++ b/v2/src/jekyll-migration/README.md
@@ -1,0 +1,5 @@
+Put pages in this directory that are a work-in-progress and should not deploy to production.
+In production, these pages will automatically produce a redirect of the form: `/<page-name>/` -> `/<page-name>.html`
+This is to facilitate the migration from Jekyll (v1) to Gatsby (v2), so that any Gatsby pages
+linking to the work-in-progress pages will appropriately redirect to their Jekyll counterparts
+during the transition.


v2/src/components/hero/hero.js
```diff
@@ -65,7 +65,7 @@ const Hero = () => {
               "See roadmap",
               "Secondary button of the hero section"
             )}
-            href={"/roadmap/"}
+            href={"/roadmap.html"}
```
jasonbcox, line 68: This is not ideal, but gets us moving so we can deploy the new homepage to production.  I'm still learning Gatsby, and it looks like there may be a way to query the config or that the page exists in order to decide whether the v1 or v2 link should be used.

In the mean time, while local testing, localhost/roadmap/ still works as expected if the roadmap page is built.

deadalnix (changes requested):
This is not going to be a maintainable solution. List of pages need to be maintained, outside of the pages themselves, in a pipeline that is not typically tested in dev, and this additionally breaks dev.

jasonbcox:
I attempted a better approach using https://www.gatsbyjs.com/plugins/gatsby-transformer-javascript-frontmatter/ to add some metadata to each page. Unfortunately, it's not possible to access this metadata using graphql in onCreatePage: https://github.com/gatsbyjs/gatsby/issues/21611

I took a hybrid approach that requires a config file for each page and screams at the developer if the config is missing, ensuring the developer is aware that the deployment metadata must be set.  This removes the mysterious build_pages.prod file that may not be obvious to someone working on one specific page.

v2/src/pages/404.js.conf
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  beta: true,
```
deadalnix, line 2: This model really is a problem for beta. What is beta supposed to be? Because right now it's going to be a random platform that doesn't look like either dev or prod.

deadalnix (changes requested):

v2/gatsby-node.js
```diff
@@ -1,4 +1,5 @@
 const locales = require("./src/i18n/locales")
+const deployPage = require("./deployPage.js")
 
 exports.onCreatePage = async ({
```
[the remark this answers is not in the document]
jasonbcox: I'm not sure what you're referring to, but if it's NODE_ENV then that's not correct. The gatsby build process overrides this and uses `production` for some parts of the build and `development` for `npm run develop`. This means you cannot override it in order to deploy dockerized dev pages, or specify a new environment that matches neither prod nor dev.

v2/src/pages/404.js.conf
jasonbcox: Ya, I see that. It's intended to be a staging area for pages that are nearing completion but not yet polished. But it doesn't really make sense since beta should just deploy all dev pages.  As we get closer to automated production deployments, beta will either reflect prod entirely or just get in the way.  I can probably just remove beta all together.

v2/src/components/hero/hero.js
jasonbcox: fixed

v2/deployPage.js
```diff
@@ -0,0 +1,22 @@
...
+
+module.exports = function(page) {
+  // `npm run develop` generates a dev-404-page. Ignore it.
+  if (page.path == '/dev-404-page/') {
+    return true
+  }
+
+  configPath = page.componentPath + '.conf'
+  if (!fs.existsSync(configPath)) {
+    throw configPath + " does not exist! Every page must have an associated configuration file."
```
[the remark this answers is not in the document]
deadalnix: Especially in dev, where the result is known and yet, the config is needed or the whole thing is made to crash.

deadalnix (changes requested):
Jason you will need to rework this. This is nowhere close to where it should be.
1/ It is cumbersome, forcing anyone and everyone to create a config file before getting started doing anything.
2/ It doesn't even work because one cannot change the config and have things work.

You need to come up with a solution that actually solves the problem and "just works".

jasonbcox:
This new approach solves the clunkiness and "just works" out of the box. This is a huge upside, but comes with some negatives: it moves `src/pages` to `src/v2pages` to avoid Gatsby's automatic page creation (which apparently cannot be disabled) and in so doing, makes the page creation a bit more complex since we can no longer rely on Gatbsy to automatically generate page paths either.

v2/src/v2pages/index.js
```diff
@@ -13,6 +13,10 @@ import HomeLargeFeature from "components/home/home-large-feature"
 import Arrow from "assets/icons/arrow.svg"
 import SingleButtonCTA from "components/CTAS/single-button-cta"
 
+export const frontmatter = {
+  prod: true,
+}
```
deadalnix, line 18: Either this is redundant

v2/src/v2pages/roadmap.js
```diff
@@ -9,6 +9,11 @@ import Rocket from "assets/icons/rocket.svg"
 import Ribbon from "assets/icons/ribbon.svg"
 import Logo from "assets/images/bitcoin-cash-logo-medium.png"
 
+export const frontmatter = {
+  prod: false,
+  v1path: '/roadmap.html',
+}
```
deadalnix, line 15: Or this is is. I'd say this is, at least for as long as the transition is ongoing.

v2/gatsby-node.js
```diff
@@ -1,31 +1,89 @@
...
+    )
+  })
+}
 
 exports.onCreatePage = async ({
   page,
   actions: { createPage, deletePage },
 }) => {
   // Delete the original page (since we are gonna create localized versions of it)
   await deletePage(page)
```
deadalnix, line 88: If this exist and is not deleting all the pages, then we can safely assume this is run on pages that are generated by gatsby itself, but not the ones we generate via createPages . It should therefore be possible to have both go over the pages folder.

v2/gatsby-node.js
```diff
@@ -1,31 +1,89 @@
...
+              toPath: localizedv1Path,
+              redirectInBrowser: true,
+              isPermanent: false,
+            })
+          }
+        }
+
+        // Create the page as we'd normally do via onCreatePage
+        const component = path.resolve(node.absolutePath)
+        console.log(component)
```
deadalnix, line 68: Remove

v2/gatsby-node.js
```diff
@@ -1,31 +1,89 @@
...
-        ...page,
-        path: localizedPath,
-        context: {
-          ...page.context,
-          originalPath,
-          locale,
-        },
-      })
-    })
-  )
```
deadalnix, line 30: You'd be better off keeping this as this as it is more generic. Just extract it in a function and reuse that function where needed.

v2/gatsby-config.js
```diff
@@ -29,6 +30,13 @@ module.exports = {
         path: "src/assets/images",
       },
     },
```
deadalnix, line 32: This is not used. You can replace it instead of adding a second import of the same module.

v2/gatsby-node.js
```diff
@@ -1,31 +1,89 @@
...
+          childJavascriptFrontmatter {
+            frontmatter {
+              prod
+              v1path
+            }
+          }
+        }
+      }
+    }
+  `)
```
deadalnix, line 24: Make sure this doesn't break hot reload. Full reload must happen when the frontmatter are changed, but not when other part of the page are.

v2/gatsby-node.js
```diff
@@ -1,31 +1,89 @@
...
+
+    // Create one page for each locale
+    Promise.all(
+      Object.entries(locales).map(async ([locale, params]) => {
+        let localizedPath = originalPath
+        if (params.slug) {
+          localizedPath = "/" + params.slug + localizedPath
+        }
+
+        if (process.env.GATSBY_APP_ENV == "prod") {
```
deadalnix, line 44: When you have a giant if that go over the whole end of the function, it's a good tell that an early return was preferable.

deadalnix (changes requested):

deadalnix:
This new approach is more promising.

v2/gatsby-node.js
jasonbcox: Hot reload of the rest of the page is unaffected.

v2/gatsby-node.js
jasonbcox: This actually doesn't cover the whole rest of the function. But considering that is not obvious in review, I'll see if I can clarify this code path.

v2/src/v2pages/index.js
jasonbcox: This explicitly needs to be true since the homepage will be deployed to prod.

v2/src/v2pages/roadmap.js
jasonbcox: This is also not redundant because without this frontmatter, `/roadmap/` will be a dead link from the homepage. We need it to redirect if the homepage is to be deployed on its own.  If anything, we're missing frontmatter on other pages since John is in the process of adding others.

v2/gatsby-node.js
jasonbcox: You're right. I intended to come back to this but it slipped my mind.

v2/gatsby-node.js
```diff
@@ -29,3 +29,90 @@ exports.onCreatePage = async ({
...
+        return false
+      }
+
+      const frontmatter = node.childJavascriptFrontmatter.frontmatter
+      if (frontmatter.prod) {
+        return true
+      }
+      if (!frontmatter.v1path) {
+        return false
+      }
```
deadalnix, line 90: None of the logic so far belong here as it is locale agnostic.

v2/gatsby-node.js
```diff
@@ -29,3 +29,90 @@ exports.onCreatePage = async ({
...
+      if (localizationParams.slug) {
+        localizedv1Path = '/' + localizationParams.slug + localizedv1Path
+      }
+      createRedirect({
+        fromPath: localizedPath,
+        toPath: localizedv1Path,
+        redirectInBrowser: true,
+        isPermanent: false,
+      })
+      return false
```
deadalnix, line 104: It does looks like you are trying to do something different than creating the page on each locale here. That really doesn't belong in a callback passed to the function to create all locales. You don't even reuse the previous slug computation.

v2/gatsby-node.js
deadalnix: I'll assume this is possible and needs to happen.

v2/src/v2pages/roadmap.js
deadalnix: You've once again setup things in a way where one has to do something for things to work as expected.

deadalnix (changes requested):

v2/gatsby-node.js
jasonbcox: It's not locale agnostic. We need redirects for the same page in each locale.

v2/gatsby-node.js
jasonbcox: See `fromPath: localizedPath`

v2/gatsby-node.js
jasonbcox: Sorry, I didn't reply to the last part. Having both go over `src/pages` does not work as intended.  `onCreatePage` is called after `createPages`, so there is no way for `createPages` to cleanup pages that we no longer want to be served.  And if we allow creation of both the pages AND redirects, it also does not work. The site only serves up the pages and does not execute the redirects as expected.

This similar issue was brought up and the Gatsby team recommends using `createPages`: https://github.com/gatsbyjs/gatsby/issues/15866 The OP also points out a workaround at the bottom of the convo, but IMO it makes too many assumptions about internal Gatsby behavior to be robust.

v2/src/v2pages/roadmap.js
jasonbcox: In dev, this will work as expected with or without frontmatter. This is only necessary if other v2 pages that are in prod have links to this page.

v2/gatsby-node.js
deadalnix: Well, if you are going to put things in another folder, then why not keep the pages by default and put the pages in dev in this new folder? That removes the need for a ton of scaffolding.

v2/src/v2pages/roadmap.js
deadalnix: Things that work as expected in dev and break in prod are not exactly the pinnacle of solid engineering.

deadalnix (changes requested):

v2/gatsby-node.js
```diff
@@ -29,3 +29,106 @@ exports.onCreatePage = async ({
...
+    // Create redirects to v1 pages if they're defined in the page frontmatter
+    createPageForAllLocales(createPage, page, (localizedPath, localizationParams) => {
+      // Create all pages in non-production environments
+      if (process.env.GATSBY_APP_ENV != "prod") {
+        return true
+      }
+
+      // By default, assuming paths of the form `/path/` transform to v1 paths
+      // of the form `/path.html`
+      let redirectPath = originalPath
```
deadalnix, line 95: let redirectPath = originalPath.replace(/\/$/, "") + ".html"

v2/gatsby-node.js
```diff
@@ -29,3 +29,106 @@ exports.onCreatePage = async ({
...
+      if (node.childJavascriptFrontmatter) {
+        const frontmatter = node.childJavascriptFrontmatter.frontmatter
+        if (frontmatter.v1pathOverride) {
+          redirectPath = frontmatter.v1pathOverride
+        }
+      }
+
+      // Localize the v1 redirect path
+      if (localizationParams.slug) {
+        redirectPath = '/' + localizationParams.slug + redirectPath
```
deadalnix, line 111: When you wrote this you should have realized the design was fucked up.

deadalnix (changes requested):

v2/gatsby-node.js
```diff
@@ -29,3 +29,106 @@ exports.onCreatePage = async ({
     })
   )
 }
+
+exports.createSchemaCustomization = ({ actions }) => {
+  const { createTypes } = actions
+
+  // Explicitly define the frontmatter schema, since it's possible the
+  // frontmatter may not be used anywhere, in which case the server will not
+  // compile due to not being able to infer the type.
```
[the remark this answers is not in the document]
jasonbcox: Gatsby complains that it cannot infer the schema since there is no frontmatter declared anywhere at the moment. One way to fix this is to introduce a dummy page that has a frontmatter set, but this seems like an unnecessary hack.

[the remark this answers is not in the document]
jasonbcox: I suppose one benefit to using a dummy is there's no extra effort to update the schema as it evolves. The dummy acts as both example and schema source of truth.

jasonbcox:
Split out some of the refactor in `gatsby-node.js` to help with the review: #580

jasonbcox:
I've removed the v1 override logic for now since we don't explicitly need it. Please review only this commit: https://github.com/bitcoincashorg/bitcoincash.org/pull/560/commits/5f90d89f51ffab772709aeae235a52b02aa5652d
Depends on #580

v2/gatsby-node.js
```diff
@@ -21,6 +22,64 @@ const forEachLocale = async (page, fn) => {
...
+    if (node.relativeDirectory) {
+      originalPath = '/' + node.relativeDirectory + originalPath
+    }
+
+    let pageFn = async (page, context, localizationParams, localizedPath) => {
+      createPage({
+        ...page,
+        path: localizedPath,
+        context,
+      })
```
deadalnix, line 56: This is duplicated from onCreatePage . Just create two functions instead, one that create redirect for all locales, and one that create pages for all locales.

deadalnix (commented):
Don't move all the pages to the migration folder. Only the one that are in unfinished state.

deadalnix (changes requested):
I meant to request changes

v2/gatsby-config.js
```diff
@@ -25,8 +25,8 @@ module.exports = {
     {
       resolve: "gatsby-source-filesystem",
       options: {
-        name: "images",
-        path: "src/assets/images",
+        name: "pages-being-migrated-from-v1",
```
deadalnix, line 28: That is certainly more explicit, but there is probably something more elegant to be found. After all, something that cannot be given a name usually means that there is a design problem.

deadalnix (changes requested):
I think you shouldn't move pages there. Discuss things with @johnkuney and figure out which one need to be moved. In the meantime, find a better name for the folder and move forward with that patch.

v2/gatsby-config.js
jasonbcox: This is why I liked `pages-dev` better. It's a natural shortening of `pages-in-development` which is exactly what they are:
* The pages do not render in production.
* There is a default redirect behavior that is beneficial for the v1 migration, but this isn't necessarily relevant for all pages in development.

But seeing as you don't think that name is appropriate with respect to the migration, I'm at a loss for a more elegant name.

jasonbcox:
* Moved pages out of the migration directory. Will coordinate with John on what belongs there in another PR.
* Prettier-formatted.
* Added a README.

v2/gatsby-config.js
```diff
@@ -25,8 +25,8 @@ module.exports = {
...
+        path: "src/pages-being-migrated-from-v1",
```
deadalnix, line 29: Call it jekyll-migration

deadalnix (changes requested):

jasonbcox:
Renamed to jekyll-migration

PR #565, opened 2020-08-28
Getting started content page

johnkuney:
Adding page for getting started content
- made reusable components for the side bar and main tile as the other pages uses these as well


v2/src/components/large-tile/large-tile.js
```diff
@@ -0,0 +1,62 @@
...
+                href={"/faq.html"}
+              />
+            </div>
+          </div>
+        </Col>
+      </>
+    )
+  }
+  return (
+    <>
```
deadalnix, line 41: No need for empty jsx

v2/src/components/large-tile/large-tile.js
```diff
@@ -0,0 +1,62 @@
...
+                <h4>{item.title}</h4>
+                <p>{item.paragraph}</p>
+              </>
+            )
+          })}
+        </div>
+      </Col>
+    </>
+  )
+}
```
deadalnix, line 60: There is a lot of duplication between the two codepath. You can put the content that is inside and change depending on the presence of a video in local variables.

v2/src/components/sidebar/sidebar.js
```diff
@@ -0,0 +1,30 @@
+import React from "react"
+import { Row, Col } from "react-bootstrap"
+import Sidebarstyle from "./sidebar.module.css"
+
+const Sidebar = ({ content, title }) => {
+  return (
+    <>
```
deadalnix, line 7: No need for the empty JSX.

deadalnix (changes requested):
Minor changes, but looks good overall.

v2/src/components/large-tile/large-tile.js
johnkuney: fixed

v2/src/components/large-tile/large-tile.js
johnkuney: Good call, condensed it some

v2/src/components/sidebar/sidebar.js
johnkuney: fixed

v2/src/components/sidebar/sidebar.js
```diff
@@ -0,0 +1,28 @@
...
+import { Row, Col } from "react-bootstrap"
+import Sidebarstyle from "./sidebar.module.css"
+
+const Sidebar = ({ content, title }) => {
+  return (
+      <Col md={4}>
+        <div className={Sidebarstyle.card}>
+          <h3>{title}</h3>
+          {content.map((item, i) => {
+            return (
```
deadalnix, line 11: When the lambda only return, you can `content.map((item, i) => returnValue`

v2/src/pages/getting-started.js
```diff
@@ -0,0 +1,109 @@
...
+          <Sidebar
+            content={sideBarContent}
+            title={fbt(
+              "Learn all about Bitcoin Cash",
+              "sidebar title 'getting started' page"
+            )}
+          />
+          <LargeTile
+            content={largeTileContent}
+            video={true}
```
deadalnix, line 100: This seems to be the only instance where this component is used. Are there others planned? If not, making a component out of it seems like it not necessary. In general, it also beg the question of have tweaks and nobs which are not used.

v2/src/pages/getting-started.js
```diff
@@ -0,0 +1,109 @@
...
+      subtitle: fbt(
+        "What is peer to peer electronic cash?",
+        "Getting started page subtitle"
+      ),
+      title: fbt("Getting Started", "Getting started page title"),
+      paragraph: fbt(
+        "Peer to peer (P2P) electronic cash is simply described as online money sent from one person to another without the need for a trusted third-party. As described in the original Bitcoin whitepaper by Satoshi Nakamoto, P2P cash makes use of digital signatures as part of the solution, but the main benefits are lost if a trusted third party is still required to prevent fraud. This makes P2P cash a trustless and safe way to transact without the need of intermediaries.",
+        "Getting started page main paragraph"
+      ),
+    },
```
deadalnix, line 83: Does it ever make sense that there would be more elements in there? If not then the whole design is a bit strange. It's unclear to me what the LargeTile abstraction provides. It doesn't seems seems like there is a clear boundary between what belong here and what belond in large tile because it doesn't really have a well defined role.

If its role is just to create a large tile, like its name indicate, then why does it know about the video or other stuffs for instance?

deadalnix (changes requested):

v2/src/components/sidebar/sidebar.js
johnkuney: Sorry not sure what you mean here

v2/src/pages/getting-started.js
johnkuney: Yes it will be on like 5 pages. The different tweaks are just because of the current legacy content. One of them has a video and the rest an image. And same for the button, only one has a button

v2/src/pages/getting-started.js
johnkuney: My main reasoning for that abstraction was to make the code more readable/easy to add content. I wanted to separate out the long text items from the LargeTile call, and like I said above the video and button specification is because only one of the pages will have that currently. Maybe an overcomplicated solution, but it will be used on five pages and now can just plug in the text and image for the rest of them, and there wont need to be a seperate CSS module for each page, it can all be contained in the large tile component

johnkuney: and yes I guess this page will have more content than the others https://www.bitcoincash.org/buy-bitcoin-cash.html and if we want to add more paragraphs or text in the future could be handy...

v2/src/components/sidebar/sidebar.js
deadalnix: See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

Notably the exemple:

```
materials.map(material => material.length)

// And not
materials.map(material => {
    return material.length
})
```

The second option is correct too, but a bit clumsy.

v2/src/pages/getting-started.js
deadalnix: Then the simplest path forward seems to be to create a component and pass either the image or video as children. If there is a common pattern that emerges for children, nothing prevents us from adding another component that specialize for this pattern of children.

v2/src/pages/getting-started.js
deadalnix: ok.

v2/src/components/sidebar/sidebar.js
johnkuney: okay for sure reformatted that

v2/src/pages/getting-started.js
johnkuney: Yeah thats cleaner, made the update

v2/src/pages/getting-started.js
```diff
@@ -0,0 +1,117 @@
...
+
+  const largeTileContent = [
+    {
+      subtitle: fbt(
+        "What is peer to peer electronic cash?",
+        "Getting started page subtitle"
+      ),
+      title: fbt("Getting Started", "Getting started page title"),
+      paragraph: fbt(
+        "Peer to peer (P2P) electronic cash is simply described as online money sent from one person to another without the need for a trusted third-party. As described in the original Bitcoin whitepaper by Satoshi Nakamoto, P2P cash makes use of digital signatures as part of the solution, but the main benefits are lost if a trusted third party is still required to prevent fraud. This makes P2P cash a trustless and safe way to transact without the need of intermediaries.",
```
deadalnix, line 80: This is quite long. If `npm run format` is unable to break it, I would advice to use a `<fbt>` tag instead.

v2/src/pages/getting-started.js
```diff
@@ -0,0 +1,117 @@
...
+            buttontext={fbt("Learn more", "learn more button")}
+            href={"/faq.html"}
+          >
+            <iframe
+              src="https://www.youtube-nocookie.com/embed/OE3QTbgh-p8?rel=0"
+              frameborder="0"
+              allow="autoplay; encrypted-media"
+              allowfullscreen
+              title="What is Bitcoin Cash"
+            ></iframe>
```
deadalnix, line 109: Can't a self closing tag be used here?

deadalnix (changes requested):
This is mostly good. A few details to polish and this is good to go.

PR #567, opened 2020-08-28
Tiles component

johnkuney:
Making tiles component for listing exchanges, wallets, services etc


v2/src/components/tiles/tiles.js
```diff
@@ -0,0 +1,28 @@
...
+import tilesStyles from "./tiles.module.css"
+
+const TilesComponent = ({ tiles, md }) => {
+  return (
+    <Container>
+      <Row>
+        {tiles.map(item => {
+          return (
+            <Col md={md} className={tilesStyles.tile}>
+              <a
```
deadalnix, line 12: You might want to use the Link tag. It just decay to a when appropriate.

PR #571, opened 2020-09-01
Featured tiles

johnkuney:
adding a featured tiles component for use on pages with a list of featured items


v2/src/components/featured-tiles/featured-tiles.js
```diff
@@ -0,0 +1,23 @@
...
+import tilesStyles from "./featured-tiles.module.css"
+import Link from "global/link.js"
+
+const FeaturedTiles = ({ tiles, md }) => {
+  return (
+    <Container>
+      <Row className={tilesStyles.tilesectionpadding}>
+        {tiles.map(item => (
+          <Col md={md} className={tilesStyles.tile}>
+            <Link to={item.link} target="_blank" rel="noreferrer">
```
jasonbcox, line 12: Why set a target? Won't this open a new tab?

PR #574, opened 2020-09-01
Translation numbering fix

johnkuney:
taking the numbers out of the fbt string on the start-here page to avoid messing with current translation


v2/src/pages/start-here.js
```diff
@@ -41,8 +41,9 @@ const StartHerePage = () => {
                 </p>
 
                 <p className={S.cardTitle}>
+                  1.&nbsp;
```
deadalnix, line 44: &nbsp; is an unbreakable space. I don't think this is what we want here.

deadalnix (changes requested):

v2/src/pages/start-here.js
johnkuney: Okay added a {" "} instead

PR #575, opened 2020-09-02
Wallets page

johnkuney:
Adding wallets page. Had to tweak some of the components to improve styling, and all of the images for the different wallets are being pulled from the static file in the root to avoid having to import each image in the page


v2/src/pages/wallets.js
```diff
@@ -0,0 +1,347 @@
...
+        </fbt>
+      ),
+    },
+  ]
+
+  const wallets = [
+    {
+      link: "https://wallet.bitcoin.com",
+      alt: "Bitcoin.com Wallet",
+      img: "/images/wallets/bitcoindotcom.png",
```
deadalnix, line 116: Given the choice, it is better to use import rather than static assets, so that webpack can do its magic.

v2/src/pages/wallets.js
```diff
@@ -0,0 +1,347 @@
...
+          Hardware{" "}
+          <fbt desc="the word 'Wallets' on the wallets page titles">
+            Wallets
+          </fbt>
+        </h2>
+        <Tiles tiles={hardwareWallets} md={4} />
+      </Container>
+      <Container>
+        <h2 className="centerh2">
+          Paper{" "}
```
deadalnix, line 336: Desktop/Hardware &al need to be translated - and in fact, re on the original website, so if you do it the same way, they should translate automatically here once transifex figure it out.

It is usually not possible to translate one word at once, because language do not have 1:1 mapping between words.

deadalnix (changes requested):

v2/src/pages/wallets.js
johnkuney: Yeah I guess so. Got them imported

v2/src/pages/wallets.js
johnkuney: Okay for sure. Yeah its weird I did it like that because thats how it is on legacy site in that particular area. They only have a translation string for the word 'wallet' and then the adjectives are just left in english. But then I just saw above it in the side bar the whole term is translated. Anyway fixed it here and it should pick up the translation from legacy site

v2/src/pages/wallets.js
```diff
@@ -0,0 +1,362 @@
...
+import Strongcoin from "assets/images/wallets/strongcoin.png"
+import Ifwallet from "assets/images/wallets/ifwallet.png"
+import Gemini from "assets/images/wallets/gemini.png"
+import Exodus from "assets/images/wallets/exodus.png"
+import Badger from "assets/images/wallets/badger.jpg"
+import Ledger from "assets/images/wallets/ledger.png"
+import Trezor from "assets/images/wallets/trezor.png"
+import Coldlar from "assets/images/wallets/coldlar.png"
+import Cashaddress from "assets/images/wallets/cashaddress.png"
+import Bitcoincashnotes from "assets/images/wallets/bitcoincashnotes.png"
```
deadalnix, line 32: So, this is clearly not going to scale or be maintainable in anyway. What about this:

1. Create a 'src/partners/wallets' folder
2. create a js file for each wallet that default export the following datastructure:
```
    {
      link: ...,
      alt: ...,
      img: ...,
      android: ...,
      ios: ...,
      windows: ...,
    },
```
3. Store the images next to the js file.
4. Use gatsby-source-filesystem to enable getting the list of all the wallets.
5. In this file, use a graphql query to query the list of wallets.
6. Go over all the wallet and construct the page.

Now we can add a new file to add a wallet, or remove one.

deadalnix (changes requested):

v2/src/pages/wallets.js
johnkuney: Sorry may need some help with that idea. I follow the logic but not entirely sure how to execute....Was able to get a folder setup and add a gatsby-source-filesystem option to the config so it can be queried , but not exactly sure on the rest of the steps....

deadalnix: OK, I'm not gatsby expert, but I will give it a go.

PR #579, opened 2020-09-03
Add one file per wallet

deadalnix:
This makes it easier to add/remove/edit wallets we support.


johnkuney:
This looks great! Yeah would've taken me a bit to get to this solution. Only comment I would say is if we should put the images in partners/wallets/images instead just to clean up the directory a bit? Not a big deal either way, and I'm happy to reaarange and update the imports later in another PR if you like

v2/src/pages/wallets.js
```diff
@@ -275,11 +91,13 @@ const WalletsPage = () => {
...
-  let androidWallets = filterWalletTypes(wallets, "android")
-  let iosWallets = filterWalletTypes(wallets, "ios")
-  let desktopWallets = filterWalletTypes(wallets, "windows")
-  let hardwareWallets = filterWalletTypes(wallets, "hardware")
-  let paperWallets = filterWalletTypes(wallets, "paper")
+  const wallets = Object.values(AllWallets).map(getWallet => getWallet())
+  const featuredWallets = filterWalletTypes(wallets, "featured")
+  const androidWallets = filterWalletTypes(wallets, "android")
+  const iosWallets = filterWalletTypes(wallets, "ios")
+  const desktopWallets = filterWalletTypes(wallets, "windows")
```
johnkuney, line 98: Also here I had realized all of the desktop wallets were on all of the operating systems so just stopped at windows and applied it to each tab. Do you think we should specify each one now with linux, mac, windows or change the attribute to desktop, and I guess update if there is ever a mac only desktop wallet or something?

deadalnix: Let's fix it in another patch.

johnkuney (approved):

PR #580, opened 2020-09-03
Refactor sharable logic in gatsby-node

jasonbcox:
This is in preparation for new `createPages` logic that will generate some pages and/or redirects based on frontmatter. `createPages` can share some of the localization logic used by `onCreatePage`, so this patch refactors to make it available.

There is no change in behavior.

### DIFF
--- a/v2/gatsby-node.js
+++ b/v2/gatsby-node.js
@@ -1,5 +1,26 @@
 const locales = require("./src/i18n/locales")
 
+// Execute some function for each locale of a given page
+const forEachLocale = async (page, fn) => {
+  await Promise.all(
+    Object.entries(locales).map(async ([locale, localizationParams]) => {
+      const originalPath = page.path
+      const context = {
+        ...page.context,
+        originalPath,
+        locale,
+      }
+
+      let localizedPath = originalPath
+      if (localizationParams.slug) {
+        localizedPath = "/" + localizationParams.slug + localizedPath
+      }
+
+      await fn(page, context, localizationParams, localizedPath)
+    })
+  )
+}
+
 exports.onCreatePage = async ({
   page,
   actions: { createPage, deletePage },
@@ -8,24 +29,11 @@ exports.onCreatePage = async ({
   await deletePage(page)
 
   // Create one page for each locale
-  await Promise.all(
-    Object.entries(locales).map(async ([locale, params]) => {
-      const originalPath = page.path
-
-      let localizedPath = originalPath
-      if (params.slug) {
-        localizedPath = "/" + params.slug + localizedPath
-      }
-
-      await createPage({
-        ...page,
-        path: localizedPath,
-        context: {
-          ...page.context,
-          originalPath,
-          locale,
-        },
-      })
+  await forEachLocale(page, (page, context, localizationParams, localizedPath) => {
+    createPage({
+      ...page,
+      path: localizedPath,
+      context,
     })
-  )
+  })
 }


v2/gatsby-node.js
```diff
@@ -1,31 +1,44 @@
...
+async function createPageForAllLocales(createPage, page, createFn) {
   await Promise.all(
     Object.entries(locales).map(async ([locale, params]) => {
       const originalPath = page.path
+      const pathWithLocale = localizePath(originalPath, params)
 
-      let localizedPath = originalPath
-      if (params.slug) {
-        localizedPath = "/" + params.slug + localizedPath
+      if (createFn) {
```
deadalnix, line 18: No need to branch here.

v2/gatsby-node.js
```diff
@@ -1,31 +1,44 @@
...
+  return path
+}
 
-  // Create one page for each locale
+// createFn is an optional override of the default createPage behavior
+async function createPageForAllLocales(createPage, page, createFn) {
   await Promise.all(
     Object.entries(locales).map(async ([locale, params]) => {
       const originalPath = page.path
+      const pathWithLocale = localizePath(originalPath, params)
```
deadalnix, line 16: That would be the `loalizedPath` ?

v2/gatsby-node.js
```diff
@@ -1,31 +1,44 @@
...
-}) => {
-  // Delete the original page (since we are gonna create localized versions of it)
-  await deletePage(page)
+// Prepend the localization slug, if there is one (English does not have one)
+function localizePath(path, params) {
+  if (params.slug) {
+    path = "/" + params.slug + path
+  }
+  return path
+}
```
deadalnix, line 9: It seems doubtful that this is useful on its own. This most likely translate a design problem in what you got coming down the road.

v2/gatsby-node.js
```diff
@@ -1,31 +1,44 @@
...
 
-  // Create one page for each locale
+// createFn is an optional override of the default createPage behavior
+async function createPageForAllLocales(createPage, page, createFn) {
```
deadalnix, line 12: Use the lambda style syntax like in the rest of the codebase.

deadalnix (changes requested):
This is a good idea to start to extract the components that work on their own.

I think think you current decoupage illustrate pretty well why your previous approach was poorly structured. Especially the branching in the middle of `createPageForAllLocales` . I'm fairly confident that you only felt the need to extract `localizePath` as a workaround for this choice.

v2/gatsby-node.js
jasonbcox: I need a way to localize the v1 path override. An alternative would be to pass a localization prefix to createFn, but this makes a stronger assumption that localized paths will always be prefix-only.

deadalnix: You don't if you start from the localized path remove the trailing slash and add ".html"

v2/gatsby-node.js
deadalnix: This still does not make sense.

deadalnix (changes requested):

v2/gatsby-node.js
jasonbcox: ...localize the v1 path **override**, not the default behavior.

v2/gatsby-node.js
jasonbcox: It looks too similar to the function name `localizePath` so I renamed it to another appropriate name.
Alternative is `const localizedPath = addLocaleToPath(...)`

jasonbcox: This naming is actually strictly more correct, so let's go with that.

v2/gatsby-node.js
deadalnix: Do we actually need a v1 path override? Can we add it if/when we need it?

deadalnix: This ^

deadalnix (changes requested):

v2/gatsby-node.js
jasonbcox: We can add it when needed. The whole point of adding it up-front was to ensure low-friction for fixing routing issues that may appear in production.

deadalnix: That's also how you end up with this marvel: https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/aop/framework/AbstractSingletonProxyFactoryBean.html

You'll note that it makes things easier if things evolve in a way you expected, but also makes things worse when thing evolve in a way you did not expect.

PR #584, opened 2020-09-05
wallet page tweaks

johnkuney:
specifying operating systems for the desktop wallets and making the sidebar anchor link to corresponding wallet sections

### DIFF
--- a/v2/src/components/sidebar/sidebar.js
+++ b/v2/src/components/sidebar/sidebar.js
@@ -7,23 +7,9 @@ const Sidebar = ({ content, title, link }) => {
     <Col md={4}>
       <div className={Sidebarstyle.card}>
         <h3>{title}</h3>
-        {link &&
-          content.map(item => (
-            <Row className={Sidebarstyle.cardIconRow}>
-              <a href={item.link}>
-                <Col xs={3} className={Sidebarstyle.cardIcon}>
-                  {item.icon}
-                </Col>
-                <Col xs={9} className={Sidebarstyle.cardIconText}>
-                  <h4>{item.title}</h4>
-                  <p>{item.paragraph}</p>
-                </Col>
-              </a>
-            </Row>
-          ))}
-        {!link &&
-          content.map(item => (
-            <Row className={Sidebarstyle.cardIconRow}>
+        {content.map(item => {
+          let tile = (
+            <>
               <Col xs={3} className={Sidebarstyle.cardIcon}>
                 {item.icon}
               </Col>
@@ -31,8 +17,13 @@ const Sidebar = ({ content, title, link }) => {
                 <h4>{item.title}</h4>
                 <p>{item.paragraph}</p>
               </Col>
-            </Row>
-          ))}
+            </>
+          )
+          if (item.link) {
+            tile = <a href={item.link}>{tile}</a>
+          }
+          return <Row className={Sidebarstyle.cardIconRow}>{tile}</Row>
+        })}
       </div>
     </Col>
   )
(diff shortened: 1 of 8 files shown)


v2/src/components/sidebar/sidebar.js
```diff
@@ -2,14 +2,28 @@ import React from "react"
...
     <Col md={4}>
       <div className={Sidebarstyle.card}>
         <h3>{title}</h3>
-        {content.map((item, i) => {
-          return (
-            <Row key={i} className={Sidebarstyle.cardIconRow}>
+        {link &&
+          content.map(item => (
+            <Row className={Sidebarstyle.cardIconRow}>
+              <a href={item.link}>
```
deadalnix, line 13: This seem to be the only difference between the two. You can make the function in the map a bit simpler instead.

```
content.map(item => {
    let tile = (<>
                <Col xs={3} className={Sidebarstyle.cardIcon}>
                  {item.icon}
                </Col>
                <Col xs={9} className={Sidebarstyle.cardIconText}>
                  <h4>{item.title}</h4>
                  <p>{item.paragraph}</p>
                </Col>
    </>)

    if (item.link) {
      tile = <a href={item.link}>{tile}</a>
    }

    return             <Row className={Sidebarstyle.cardIconRow}>{tile}</Row>
  }
}
```

v2/src/pages/wallets.js
```diff
@@ -110,6 +116,7 @@ const WalletsPage = () => {
               "What kind of wallet should I use?",
               "sidebar title 'wallets' page"
             )}
+            link={true}
```
deadalnix, line 119: This is not necessary.

v2/src/pages/wallets.js
```diff
@@ -125,7 +132,7 @@ const WalletsPage = () => {
         <FeaturedTiles tiles={featuredWallets} md={6} />
       </Container>
       <Container>
-        <h2 className="centerh2">
+        <h2 className="centerh2" id={"mobile-wallets"}>
```
deadalnix, line 135: You don't need curly braces.

v2/src/pages/wallets.js
```diff
@@ -140,33 +147,33 @@ const WalletsPage = () => {
         </Tabs>
       </Container>
       <Container>
-        <h2 className="centerh2">
+        <h2 className="centerh2" id={"desktop-wallets"}>
```
deadalnix, line 150: dito

deadalnix (changes requested):

deadalnix:
In general, you want to split your patches so that they do one thing. This does two (anchor links and proper OS support for desktop wallets) meaning both will have to be good before any land.

v2/src/components/sidebar/sidebar.js
johnkuney: Ah nice, yeah thats a cool way to do it. Updated

v2/src/pages/wallets.js
johnkuney: removed

v2/src/pages/wallets.js
johnkuney: fixed

v2/src/pages/wallets.js
johnkuney: fixed

v2/src/components/sidebar/sidebar.js
```diff
@@ -2,23 +2,27 @@ import React from "react"
 import { Row, Col } from "react-bootstrap"
 import Sidebarstyle from "./sidebar.module.css"
 
-const Sidebar = ({ content, title }) => {
+const Sidebar = ({ content, title, link }) => {
```
deadalnix, line 5: link is not used anymore

deadalnix (changes requested):
You missed a detail, but overall, LGTM

v2/src/components/sidebar/sidebar.js
johnkuney: grrr good catch. Fixed

PR #588, opened 2020-09-09
Graphics page

johnkuney:
Creating graphics page for logos and visual identity.

Given these items do not change often and when they do likely all change at once, and that the downloadable images are not shown on the page, I opted to put the images in the static folder as it is much easier to manage than importing like we're doing with exchanges etc.


v2/src/pages/graphics.js
```diff
@@ -0,0 +1,367 @@
...
+              />
+            </div>
+          </Col>
+          <Col md={6} className={S.colorctnpadding}>
+            <Row>
+              <Col xs={4} className={S.colorctn}>
+                <span
+                  className={S.boxcolor}
+                  style={{ background: "#F59332" }}
+                ></span>
```
deadalnix, line 295: You can use a self closing tag

deadalnix (approved):

deadalnix (changes requested):
You need to update the links.

v2/src/pages/graphics.js
johnkuney: Ok sure updated

johnkuney:
Okay link in menu has been updated. The only reason I left it off is because wont it cause merge conflicts if I change the header links for different PRs? Thats why I only did one of them and was planning on doing the other right after...

v2/src/pages/graphics.js
```diff
@@ -0,0 +1,367 @@
...
+import AcceptedHere from "assets/images/logos/bitcoin-cash-accepted-here.jpg"
+import AcceptedHereOrange from "assets/images/logos/bitcoin-cash-logo-accepted-here-orange.jpg"
+import Donations from "assets/images/logos/bitcoin-cash-logo-donations-accepted.jpg"
+import DonationsOrange from "assets/images/logos/bitcoin-cash-logo-donations-accepted-orange.jpg"
+import DownloadIcon from "assets/icons/download.svg"
+
+const GraphicsPage = () => {
+  return (
+    <>
+      <SEO title="Bitcoin Cash Wallets" />
```
deadalnix, line 24: ???

johnkuney: ahh sorry missed it. Will update

PR #590, opened 2020-09-09
Services page

johnkuney:
adding services page


deadalnix (changes requested):
memo.png is not used

johnkuney:
Okay removed

PR #591, opened 2020-09-09
Properly localize the upgrade date.

deadalnix:
As per title


johnkuney (approved):
Cool, looks good to me

PR #597, opened 2020-09-09
Use fbt:param in spend bitcoin cash page

deadalnix:
There is no point to try to translate domain names.


johnkuney (approved):
Cool looks good to me

PR #601, opened 2020-09-11
Update SEO

johnkuney:
Added default page description, twitter card type, twitter card image, author


v2/gatsby-config.js
```diff
@@ -1,8 +1,8 @@
 module.exports = {
   siteMetadata: {
     title: "Bitcoincash.org",
-    description: "Money built for you—not banks",
-    author: "@gatsbyjs",
+    description: "Bitcoin Cash brings sound money to the world. Merchants and users are empowered with low fees and reliable confirmations. The future shines brightly with unrestricted growth, global adoption, permissionless innovation, and decentralized development.",
```
deadalnix, line 4: This is not the right approach, as this will not be translated.

deadalnix (changes requested):
Once again, you setup yourself for problem by packaging things that are independent together. The twitter image thing is fine, but the description via metadata is not working.

v2/gatsby-config.js
deadalnix: in fact, it is not translatable.

v2/src/pages/accept-bitcoin-cash.js
```diff
@@ -90,7 +90,13 @@ const AcceptPage = () => {
 
   return (
     <>
-      <SEO title="Spend Bitcoin Cash" />
+      <SEO
+        title="Spend Bitcoin Cash"
+        description={fbt(
+          "Bitcoin Cash brings sound money to the world. Merchants and users are empowered with low fees and reliable confirmations. The future shines brightly with unrestricted growth, global adoption, permissionless innovation, and decentralized development.",
+          "Default SEO page description"
+        )}
```
deadalnix, line 98: Don't repeat the same code in 20 different places.

v2/src/pages/accept-bitcoin-cash.js
```diff
@@ -90,7 +90,13 @@ const AcceptPage = () => {
 
   return (
     <>
-      <SEO title="Spend Bitcoin Cash" />
+      <SEO
+        title="Spend Bitcoin Cash"
```
deadalnix, line 94: But you might want to consider translating titles in another patch.

v2/src/components/seo.js
```diff
@@ -94,11 +98,12 @@ function SEO({ description, lang, meta, title }) {
 SEO.defaultProps = {
   lang: `en`,
   meta: [],
-  description: ``,
+  description:
```
deadalnix, line 101: You will not be able to put it is here, because these are static values, and translation is not static.

v2/src/components/seo.js
```diff
@@ -9,23 +9,23 @@ import React from "react"
...
             title
-            description
             author
           }
         }
       }
     `
   )
 
-  const metaDescription = description || site.siteMetadata.description
```
deadalnix, line 28: So you need to resurrect this, with the default value inline, using an fbt.

deadalnix (changes requested):

PR #603, opened 2020-09-16
Price ticker update

johnkuney:
Making different currencies available in a dropdown on the price ticker

### DIFF
--- a/v2/src/components/liveprice/live-price-widget.js
+++ b/v2/src/components/liveprice/live-price-widget.js
@@ -1,49 +1,66 @@
-import React, {useState} from "react"
+import React, { useState } from "react"
 import livePricingStyles from "./live-price-widget.module.css"
 import Triangle from "assets/icons/triangle.svg"
 
-const LivePriceWidget = ({ currentUSDPrice, currentEURPrice, currentJPYPrice, currentCNYPrice }) => {
-  const [open, setOpen] = useState(false);
-  const [hide, setHide] = useState(false);
-  const [selectedCurrency, setCurrency] = useState("");
+const LivePriceWidget = ({
+  currentUSDPrice,
+  currentEURPrice,
+  currentJPYPrice,
+  currentCNYPrice,
+}) => {
+  const [open, setOpen] = useState(false)
+  const [hide, setHide] = useState(false)
+  const [selectedCurrency, setCurrency] = useState("")
   return (
     <div className={livePricingStyles.widgetContainer}>
-        <div onClick={() => setOpen(!open)} className={livePricingStyles.widget}>
-          {selectedCurrency}<span className={hide ? `${livePricingStyles.hide}` : "" }>{currentUSDPrice}</span><Triangle />
-        </div>
-  
-        {open &&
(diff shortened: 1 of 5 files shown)


v2/src/components/liveprice/live-price-widget.js
```diff
@@ -1,12 +1,67 @@
...
+
+      {open && (
+        <ul className={livePricingStyles.priceDropdownUl}>
+          <li
+            onClick={() => {
+              setOpen(!open)
+              setCurrency(currentUSDPrice)
+              setHide(true)
+            }}
+          >
```
jasonbcox, line 32: Just loop through a list of the prices so you don't have to copy-paste this code.

jasonbcox (changes requested):

v2/src/components/liveprice/live-price-widget.js
```diff
@@ -1,12 +1,67 @@
...
 
-const LivePriceWidget = ({ currentPrice, ticker, url }) => {
+const LivePriceWidget = ({
+  currentUSDPrice,
+  currentEURPrice,
+  currentJPYPrice,
+  currentCNYPrice,
+}) => {
+  const [open, setOpen] = useState(false)
+  const [hide, setHide] = useState(false)
```
deadalnix, line 12: hide is a verb. I assume you meant hidden. What does this mean for the LivePriceWidget to be hidden (or open for that matter?)

v2/src/layout/header.js
```diff
@@ -117,13 +117,19 @@ const Header = () => {
   ]
 
   const bchPriceApi =
-    "https://min-api.cryptocompare.com/data/price?fsym=BCH&tsyms=USD"
+    "https://min-api.cryptocompare.com/data/price?fsym=BCH&tsyms=USD,JPY,EUR,CNY"
```
deadalnix, line 120: It's clear you have a bunch of logic here that doesn't really belong here. All the price management business is the role of the price widget, not of the header.

v2/src/layout/header.js
```diff
@@ -173,9 +179,26 @@ const Header = () => {
...
+              currentJPYPrice={[
+                "¥",
+                currentJPYPrice,
+                <span className={headerStyles.ticker}>JPY</span>,
+              ]}
+              currentCNYPrice={[
+                "¥",
+                currentCNYPrice,
+                <span className={headerStyles.ticker}>CNY</span>,
+              ]}
```
deadalnix, line 201: When you start writing this, you should pause a second and ask yourself if that really does make sense. If there was 1000 currencies, you'd copy paste the same stuff 1000 times?

deadalnix (changes requested):
If I am to be honest, the code is pretty bad here. You did not inherit a great design to begin with, but doubling down on the mess is not good engineering.

The general currency selection doesn't make much sense. It always falls back to USD. You want it to go as follow:
 - If the user never chose, then use whatever the locale says should be used.
 - If the user chose, then use whatever the user chose, don't fall back on USD.

johnkuney:
Thanks for this feedback. I agree there is some messy stuff going on. I'll rethink my approach

PR #604, opened 2020-09-17
CSS fix on links in sidebar

johnkuney:


deadalnix:
What problem does this solve?

johnkuney:
Sorry should have left a description. Just a style patch for the links that are in the sidebar component in the getting started pages (/buy-bitcoin-cash/ ex) The text links within the paragraph needed to be specified to inline so they dont break lines

jasonbcox (changes requested):
This breaks the rendering of the sidebar container in Firefox 68.12.x and I still see breaking of some of the lines since they're way too long to fit on one line. I'm not sure why forcing them to not break is beneficial.

johnkuney:
I said the wrong word. I want the links to break. Currently they take an entire line of the paragraph like you are describing above. By making them inline they should flow with the paragraph right? I'm seeing the opposite of what youre describing with this change. Are you seeing that running this or thats what you see on the live site?

deadalnix:
Ok so your problem is that the sidebar display are defined as flex, which is good, but every child is also defined as flex, which is definitely wrong. You need to fix the sidebar CSS instead of doing this.

deadalnix:
You want to change this into `.card > a` or something like that.

johnkuney:
Okay I specified a class for the links that need flex so it doesnt get applied to the others

jasonbcox (dismissed):

deadalnix:
> 
> 
> You want to change this into `.card > a` or something like that.

deadalnix (changes requested):
You don't need to add a new class or anything. See comment I made twice now.

johnkuney:
this doesnt work for the rows that need to be links on the wallet page

v2/src/components/sidebar/sidebar.module.css
```diff
@@ -8,16 +8,17 @@
...
+.card > a,
+.card > a:visited,
+.card > a:active {
   display: flex;
   color: #444;
   transition: all 300ms;
 }
 
-.card a:hover {
+.clickableRow:hover {
```
deadalnix, line 19: Why is this clickableRow now?

johnkuney:
those are a children too and need specific style

deadalnix (changes requested):
I don't think we have clickableRow anywhere in the codebase at this time, so i don't see how that CSS element makes any sense.

v2/src/components/sidebar/sidebar.module.css
johnkuney: thats the class name

deadalnix:
This is wrong: https://github.com/bitcoincashorg/bitcoincash.org/pull/604/files#diff-c084369006e972f978373026b3dc660cR14

It defines all the childs as flew too.

johnkuney:
yes thats why I added the class before. Some of the links in the sidebar need different styling

deadalnix:
There are no other links in your patch that use the class, so I still don't see how any of this makes any sense.

johnkuney:
There are two possible types of links that need different styling in this sidebar component. 

One is the type that I made this class for. It needs display: flex some colors etc. These are the anchor links that take you to different sections found on the wallet page for example that take up a whole row.

The other type is a link within the paragraph text. These only need their color defined and the rest of the default styling is fine. So I selected them through their parent class and applied the color no class needed

deadalnix:
The anchor links are not immediate childs of .card are they?

The these anchor link cannot appear anywhere else. So why `.card > a` inadequate?

johnkuney:
The anchor links are not direct children of .card. They are children of .cardIconRow so .cardIconRow > a works

PR #608, opened 2020-09-22
Header refactor

johnkuney:
This PR looks to break out the nav items from the header component in order to simply the components and not have the header have to do too much like worry about the state of the nav

### DIFF
--- a/v2/src/layout/navbar.js
+++ b/v2/src/layout/navbar.js
@@ -84,7 +84,7 @@ const NavBar = () => {
     },
   ]
 
-  const [mobile, openMobile] = useState(false)
+  const [open, openMobileMenu] = useState(false)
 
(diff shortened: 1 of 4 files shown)


v2/src/layout/navbar.js
```diff
@@ -0,0 +1,191 @@
...
+          <span style={{ paddingLeft: "5px" }}>
+            <fbt desc="Language selector menu">Language</fbt>
+          </span>
+        </>
+      ),
+      links: languageDropdownLinks,
+    },
+  ]
+
+  const [mobile, openMobile] = useState(false)
```
deadalnix, line 87: mobile? I get what expanded means for a menu, but mobile? Does it means we can move it around?

deadalnix (dismissed):
If you are going to rename things - which is good - make sure to pick a proper name.

PR #609, opened 2020-09-22
Fix twitter card image path

johnkuney:
I realized twitter card image needs an absolute path to work. So used the static folder and referenced the url directly. 
- moved twitter image to static folder
- added bitcoincash.org url + file path to twitter image so it will work when live on bitcoincash.org


v2/src/components/seo.js
```diff
@@ -73,7 +72,7 @@ function SEO({ description, lang, meta, title, twitter_image }) {
         },
         {
           name: `twitter:image`,
-          content: metatwitterImage,
+          content: "https://www.bitcoincash.org/images/" + metatwitterImage,
```
deadalnix, line 75: Do not bake the domain in the code

deadalnix (changes requested):
Why dos it need an absolute path to work?

johnkuney:
You'd have to ask twitter, but its stated in this thread https://twittercommunity.com/t/card-error-unable-to-render-or-no-image-read-this-first/62736 third from last bullet

v2/src/components/seo.js
johnkuney: For sure. What about if we define it in the gatsby config meta data and bring it in like that?

v2/gatsby-config.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
   siteMetadata: {
     title: "Bitcoincash.org",
     author: "Bitcoincash.org",
+    siteURL: "https://www.bitcoincash.org/",
```
deadalnix, line 5: I don't see how putting the absolute url somewhere else solves anything.

deadalnix (changes requested):
So, I've loaded the site from the branch in your repo. I can't see anything different in either chrome or firefox.

It remains unclear what problem this is solving if any, and, in any case, the solution needs to adjust to the source from which the content is served.

johnkuney:
So this problem is as of now this twitter card image is not working with twitter. You can test it out here https://cards-dev.twitter.com/validator

According to their docs the image needs an absolute path to work

So need to define the absolute path. I agree having it adjust to the serving source is ideal. Will see if I can make that happen

johnkuney:
hmm may be stuck with it. Seems gatsbys own docs recomend defining the domain since gatsby cant really see it 
Screen Shot 2020-09-22 at 4 04 54 PM

deadalnix:
Alright, but then it mean that passing the twitter image as argument is not going to work, so we got to remove that option.

v2/src/components/seo.js
deadalnix: OK so it seems like there is no good way to do this. Then I guess we can go with this.

Can you set the images/ as part of metatwitterImage rather than prebaked?

deadalnix (changes requested):
One small change and this can be merged.

johnkuney:
Okay made that edit. Was that other comment still relevant? It looks to me we can still use the argument. Will just need to pass a string of the image path wanted. Seems to work in testing

PR #610, opened 2020-09-22
CSS tweaks for mobile

johnkuney:
These are a few CSS tweaks to improve the spacing on some elements in mobile view


deadalnix:
You should really explain what problem these are solving.

johnkuney:
For sure will be more specific in the future

PR #611, opened 2020-09-22
Generate v1 -> v2 redirects for all production pages

jasonbcox:
With v1 pages on their way out, we need to redirect v1 pages to v2 pages. This way, links to the old v1 routes will still work, but see the new experience.

This patch adds the necessary machinery to introduce those redirects and is testable as a stand-alone change. However, in order to push this to production, the v1 configs in nginx need to be cleaned up first. That patch will soon follow.

#### Test plan:
```
npm run develop
```
Navigate around. Pages still work as expected and v1 links such as `localhost:8000/start-here.html` redirect to their appropriate v2 counterparts.

```
mv src/pages/accept-bitcoin-cash.js src/jekyll-migration
GATSBY_APP_ENV=prod npm run develop
```
In addition to the above tested behavior, pages in the jekyll migration directory still redirect in the v2 -> v1 direction as expected.


v2/gatsby-node.js
```diff
@@ -92,7 +100,7 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
       // In production, we redirect in-development v2 pages to v1 equivalent routes.
       // In non-prod environments, we create all pages and do not redirect.
       if (process.env.GATSBY_APP_ENV == "prod") {
-        await createRedirectForEachLocale(createRedirect, page)
+        await createRedirectForEachLocale(createRedirect, page, fromHtmlPage = false)
```
deadalnix, line 103: I'm pretty sure this creates a global variable and by no mean is a named argument

v2/gatsby-node.js
```diff
@@ -35,15 +35,23 @@ const createPageForEachLocale = async (createPage, page) => {
   )
 }
 
-const createRedirectForEachLocale = async (createRedirect, page) => {
+const createRedirectForEachLocale = async (createRedirect, page, fromHtmlPage = true) => {
```
deadalnix, line 38: Boolean in API = beark.

Just pass a customized version of createRedirect that do the switcharoo and rename a few things so that they make sense.

deadalnix (changes requested):

PR #613, opened 2020-09-23
Add Jest testing skeleton and one snapshot test

jasonbcox:
This introduces a framework for writing various types of tests, as well as a very bare-bones test so that we know everything works.

Most of this is unit test scaffolding pulled from https://www.gatsbyjs.com/docs/unit-testing/

The tweaks to PrimaryButton are inspired by https://www.gatsbyjs.com/docs/testing-components-with-graphql/

#### Test Plan
```
npm install
npm run test
```
And for sanity:
```
npm run develop
```
Navigate to localhost:8000

### DIFF
--- a/v2/__mocks__/file-mock.js
+++ b/v2/__mocks__/file-mock.js
@@ -1,0 +1,1 @@
+module.exports = "test-file-stub"
--- a/v2/__mocks__/gatsby.js
+++ b/v2/__mocks__/gatsby.js
@@ -1,0 +1,26 @@
+const React = require("react")
+const gatsby = jest.requireActual("gatsby")
+
+module.exports = {
+  ...gatsby,
+  graphql: jest.fn(),
+  Link: jest.fn().mockImplementation(
+    ({
+      activeClassName,
+      activeStyle,
+      getProps,
+      innerRef,
+      partiallyActive,
+      ref,
+      replace,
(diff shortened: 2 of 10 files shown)


v2/jest/preprocess.js
```diff
@@ -0,0 +1,5 @@
+const babelOptions = {
+  presets: ["babel-preset-gatsby"],
+}
+
+module.exports = require("babel-jest").createTransformer(babelOptions)
```
deadalnix, line 5: That would be preferable if this could use the .babelrc provided. This is not going to handle fbt properly for instance.

deadalnix (changes requested):
That seems like a lot of crap to just test one button XD

Is this really how people do this in the JS ecosystem?

v2/jest/preprocess.js
jasonbcox: Apparently it uses .babelrc out of the box. Not sure why this note is hidden away in their docs, but everything appears to work just removing this file all together and getting rid of the explicit `transform:` config line.

jasonbcox:
> That seems like a lot of crap to just test one button XD
> 
> Is this really how people do this in the JS ecosystem?

Funny you say this, because this might be the best state I've seen the JS ecosystem in over the last 10 years LOL

PR #620, opened 2020-09-25
Improve link style on upgrade page

johnkuney:
The links on the upgrade page needed to look more like links
- added link icon and underline text decoration


deadalnix:
Ok, this is certainly better, but why aren't all the link looking the same to begin with? It seems odd that the link style differs for many pages.

johnkuney:
For sure. The main places I'm seeing text links are in the sidebar component and the announcement bar. Both are styled pretty similar as of now except the color which needs to change based on the background color. I personally don't really like underlined styling on non hovered inline text links which is why I left it off, but agree when a link is just floating by itself it needs some more indication. Sooo I'll go ahead a make the global default underlined and then for floating links can add the little icon. As for color I guess makes sense to have orange be the default and then will have to specify for the announcement bar because orange would be hard to read on yellow background

johnkuney:
I ended up defining the global link css as orange and underlined. The styling for other specific links on the page like the nav items, homepage sections, etc were previously relying on the the global styling but seeing as they are generally the exception I thought it made the most sense to define those specifically and make the default style as we would want inline text links to appear as those will likely be the most common.

PR #627, opened 2020-09-29
Footer update

johnkuney:
The footer was pretty basic so I wanted to fill it out to more of a traditional footer with links to pages. 
- Added bitcoin cash logo and moved copyright text up
- added three columns of page categories with links
- updated telegram logo


v2/src/layout/footer.js
```diff
@@ -11,21 +12,69 @@ import Instagram from "assets/icons/social/instagram.svg"
...
+    },
+    {
+      text: <fbt desc="Spend page footer title">Spend Bitcoin Cash</fbt>,
+      href: "/spend-bitcoin-cash/",
+    },
+    {
+      text: <fbt desc="accept page footer title">Accept Bitcoin Cash</fbt>,
+      href: "/accept-bitcoin-cash/",
+    },
+  ]
```
deadalnix, line 36: This seems to be the exact same thing as in the top menu, so you need to factor it.

deadalnix (changes requested):

v2/src/layout/footer.js
johnkuney: Okay I made some adjustments so the header and footer can pull from the same source of links. Only issue is I cannot get it to work with <fbt>? So as it is it is not ready to merge, but works without fbt. Obviously need to fix that though...

Do you have any ideas why it might be breaking with fbt items in the arrays?

v2/src/layout/footer.js
```diff
@@ -1,102 +1,109 @@
...
+          </Col>
+        </Row>
+      </Container>
+      <div>
+        <div className={S.socialBar}>
+          <a className={S.socialIcon} href="https://twitter.com/bitcoincashorg">
             <Twitter />
           </a>
-          <a className={footerStyles.socialIcon} href="https://t.me/BCHUpdates">
+          <a className={S.socialIcon} href="https://t.me/BCHUpdates">
```
deadalnix, line 82: https://t.me/bitcoincashfork might be best.

v2/src/layout/nav-items.js
```diff
@@ -0,0 +1,137 @@
...
+  if (locale.slug) {
+    link += locale.slug + "/"
+  }
+
+  return { text: locale.displayName, href: link, localize: false }
+})
+
+const communityDropdownLinks = [
+  {
+    text: "Services",
```
deadalnix, line 17: You just busted all the translations. How is that even remotely in the real of what get submitted for reviews?

deadalnix (changes requested):

v2/src/layout/nav-items.js
johnkuney: Did you see my comment above? I addressed this and asked if you had any ideas. The fbt is breaking the build when applied within that component. I'm not sure why, but will continue to troubleshoot

deadalnix: Because it is at top level and is therefore computed before the locale is setup. You need to wrap these thing into functions.

deadalnix (changes requested):

v2/src/layout/nav-items.js
johnkuney: Okay thanks for that tip think I got it

PR #635, opened 2020-10-12
[activationtestnet] add addnode= commands to ensure nodes find our infra

majcosta:
as title


jasonbcox (approved):
Good for now. We'll update the IPs once the DNS updates are ready.

PR #639, opened 2020-10-13
[activationtestnet] make example easier to read and copy

majcosta:
easier to read, easier to copy&paste and change the activation timestamp

also changed from old `-testnet` switch to newer style `-chain=test`


jasonbcox (changes requested):
This might render well, but it's difficult to update without making mistakes. There should be a way to do this with multiple `>`

jasonbcox (approved):

PR #641, opened 2020-10-15
Whitepaper page

johnkuney:
The whitepaper link in the menu was acting up because it was being treated as an internal link, when really it needs to be non gatsby link. One solution is to instead embed the pdf on a whitepaper page. I thought this could be good so the site nav is still present while viewing the pdf. 
-made whitepaper page to embed pdf
-borrowed translated versions from bitcoin.org and used slug to load translated versions based on language selected


v2/src/pages/whitepaper.js
```diff
@@ -0,0 +1,22 @@
+import React from "react"
+import fbt from "fbt"
+import SEO from "components/seo"
+import { useLocaleContext } from "i18n/provider"
+
+const Whitepaper = () => {
+  const { slug } = useLocaleContext()
+  return (
+    <>
+      <SEO title={fbt("Bitcoin Whitepaper", "Whitepaper page SEO title")} />
```
jasonbcox, line 10: The iframe will likely nuke any potential SEO value, but it doesn't hurt :P

jasonbcox (approved):

PR #645, opened 2020-10-16
Correct whitepaper file name

johnkuney:
adding condition so file name for english pdf wont end in a "-"


jasonbcox (approved):

v2/src/pages/whitepaper.js
```diff
@@ -5,12 +5,16 @@ import { useLocaleContext } from "i18n/provider"
 
 const Whitepaper = () => {
   const { slug } = useLocaleContext()
+  let prefix = ""
```
jasonbcox, line 8: Suffix, not prefix

jasonbcox (changes requested):

v2/src/pages/whitepaper.js
johnkuney: For sure, fixed.
Fyi I also added "#view=FitH" to the embed to help with making the view the fit different screen sizes

jasonbcox (approved):

PR #647, opened 2020-10-17
Whitepaper page revisions

johnkuney:
After seeing the whitepaper page in production I think the iframe approach is not ideal. Too many different browser interpretations for displaying the pdf and was seeing some issues on different mobile views that were not apparent in testing.

So now looking at having the page be a landing spot for links to the different language versions of the document.

Can maybe later (with Christables help) add some text explaining what the whitepaper is to fill out the page more, but for now this solution mitigates pdf display issues and provides different language options to the user


v2/src/pages/whitepaper.js
```diff
@@ -1,24 +1,41 @@
...
+          <PDF href="/bitcoin-fr.pdf" label="Français" />
+          <PDF href="/bitcoin-id.pdf" label="Bahasa Indonesia" />
+          <PDF href="/bitcoin-ja.pdf" label="日本語" />
+          <PDF href="/bitcoin-ko.pdf" label="한국어" />
+          <PDF href="/bitcoin-nl.pdf" label="Nederlands" />  
+          <PDF href="/bitcoin-pt-BR.pdf" label="Português Brasil" />  
+          <PDF href="/bitcoin-pt-PT.pdf" label="Português Portugal" />  
+          <PDF href="/bitcoin-ru.pdf" label="Русский" />  
+          <PDF href="/bitcoin-tr.pdf" label="Türkçe" />  
+          <PDF href="/bitcoin-zh-CN.pdf" label="简体中文" />
```
deadalnix, line 35: You want to feature the user's language and the English because it is the original.

Also, why not loop over the locales instead of doing this? This doesn't looks very maintainable.

deadalnix (changes requested):

v2/src/pages/whitepaper.js
johnkuney: Okay for sure. Made it so the user's language is first and then english, and then it maps over the other languages. 

Also made it so the display name for english says "English (Original)" instead of using the display name "English (US)" so its clear the english version is the original version

PR #660, opened 2020-11-11
Updating upgrade info

johnkuney:
Updating the info in the announcement bar and upgrade page to reflect the latest info.
- added new text
- changed banner behavior to be closed by default
- added other compatible node links 
- added share button that copies the upgrade page url to clipboard

### DIFF
--- a/v2/src/layout/announcement-bar.js
+++ b/v2/src/layout/announcement-bar.js
@@ -1,12 +1,12 @@
 import React from "react"
 import fbt from "fbt"
 import S from "./announcement-bar.module.scss"
-import PrimaryButton from "components/buttons/primary-button"
 import Link from "global/link"
 import { Accordion } from "react-bootstrap"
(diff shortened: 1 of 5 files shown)


v2/src/layout/announcement-bar.js
```diff
@@ -92,11 +92,36 @@ const CountdownClock = ({ days, hours, minutes, seconds, completed }) => {
...
+      <LinkIcon />
+      {text}
+    </Link>
+  )
+}
+
+const CopyLink = () => {
+  const el = document.createElement("textarea")
+  const notice = document.getElementById("copiedNotice")
+  el.value = "https://www.bitcoincash.org/upgrade/"
```
jasonbcox, line 107: Nit: The link could be an arg to the function to make this more useful in other places in the future. But that can wait until it's actually needed elsewhere.

jasonbcox (approved):

